PR #9, opened 2018-08-01
Feature/limit ingress networks

dav3r:
Enable limiting access to Nessus instance via trusted networks list in variables.tf (or any other supplied variables file, like production.tfvars)


jsf9k (approved):
This looks good to me!

PR #12, opened 2018-08-02
Change from Mongo 4.0.0 to Mongo 3.2.20 to match version of current C…

dav3r:
…yHy data

### DIFF
--- a/terraform/scripts/mongo_disk_setup.yml
+++ b/terraform/scripts/mongo_disk_setup.yml
@@ -11,7 +11,7 @@ mounts:
 
 bootcmd:
   # Wait for EBS disks to be attached
-  - 'while [ `lsblk | wc -l` -lt 6 ]; do echo Waiting for disks to attach...; sleep 5; done'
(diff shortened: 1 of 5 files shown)


terraform/scripts/mongo_disk_setup.yml
```diff
@@ -11,7 +11,7 @@ mounts:
 
 bootcmd:
   # Wait for EBS disks to be attached
-  - 'while [ `lsblk | wc -l` -lt 6 ]; do echo Waiting for disks to attach...; sleep 5; done'
+  - 'while [ `lsblk | grep -c " disk"` -lt 4 ]; do echo Waiting for disks to attach...; sleep 5; done'
```
jsf9k, line 14: I asked @dav3r if this change was intentional, and he said it was.  When he booted up in Jessie the root volume had two partitions, which broke the old logic.  He amended the logic to work in both cases.

jsf9k (approved):
Top notch work, @dav3r!

PR #14, opened 2018-08-02
Add Terraform configuration for BOD 18-01 scanning

jsf9k:
This isn't complete, but I'd like to go ahead and merge this and do further work in multiple pull requests.


dav3r (commented):
Lots in here, but it all made sense to me.  Keep on truckin' @jsf9k !
I expect that sometime soon, we will want to lock down public access to only allow ingress from our trusted networks (mainly for production deployments).  Similar to here: https://github.com/dhs-ncats/cyhy_amis/blob/bca515dcf927c2e0eb63ef914e2eae119f4f7914/terraform_nessus_only/vpc.tf#L155

dav3r (approved):

PR #15, opened 2018-08-03
Add VPC peering between the CyHy and BOD VPCs

jsf9k:


felddy (approved):

ansible/roles/mongo/tasks/main.yml
```diff
@@ -28,6 +28,13 @@
     state: present
     line: After=network.target cloud-config.service
 
+- name: Configure mongo service to listen on all interfaces
+  lineinfile:
+    dest: /lib/systemd/system/mongod.service
+    regexp: '^( *bindIp=.*)$'
```
jsf9k: This isn't working.  :angry:   I need to fix this before we merge.

jsf9k (commented):
Needs a little work before merging :disappointed:

jsf9k:
Building up and tearing down is complicated now that we have two Terraform projects (`terraform` and `terraform _bod`) that are interdependent.  Maybe we should move the `terraform_bod` code into the `terraform` directory.

What do you guys think?

ansible/roles/mongo/tasks/main.yml
dav3r: There are a couple other changes that I made to the mongo config file (located at /etc/mongod.conf, not /lib/systemd/system/mongod.service), so let's coordinate on those too.

dav3r:
@jsf9k , I agree that if the `terraform` and `terraform_bod` directories are now interdependent, they should be merged.

jsf9k:
@dav3r, I moved the `terraform_bod` code into the `terraform` directory (and renamed and modified some files as necessary).  Let's get those Mongo configuration changes you have in so we can merge this one!
:penguin: :tada: :trumpet: :smiling_imp: :cat2: :koala: :dolphin:

dav3r (approved):
👍 @jsf9k

PR #18, opened 2018-08-05
Provision MongoDB users via ansible

jsf9k:
* The MongoDB users are now provisioned via `ansible`.
* Our `terraform` infrastructure now destroys cleanly.
* ssh proxying inside `ansible` now works as expected.


jsf9k:
If anyone has half-finished stuff in this branch that isn't ready for merge just add a comment.  It would be nice if we could get this approved and merged before folks leave for [Lost Wages](https://media.giphy.com/media/ShZ1AHZ1AKyt2/giphy.gif), though.  That way we can all start from clean feature branches.

jsf9k:
@dav3r, if you have a chance this week try adding the line `"-vvv"` at [line 52 of `cyhy_mongo_ec2.tf`](https://github.com/dhs-ncats/cyhy_amis/blob/provision_db_users/terraform/cyhy_mongo_ec2.tf#L52) and see if the extra debugging information from `ansible` is of any help.  It acts like it can't find the `boto3` and/or `botocore` Python libraries on your system.

I'll also keep an eye out for this error as I tear down and build up in the coming days.  And I'll let you know if I think of anything else.

jsf9k:
@dav3r
For what it's worth, I just built from scratch in us-west-1 and did not have the problem you are seeing.  I think it must be something on your end, or at least something OSX-specific.  Hopefully the directions in my previous comment will shed some light on what is going on.  Maybe Ansible is using a different `python` that doesn't have `boto3` installed?

dav3r (approved):
This is fantastic work @jsf9k

PR #19, opened 2018-08-09
Fixes made while debugging BOD 18-01 scanning and report generation

jsf9k:


terraform/cyhy_mongo_ec2.tf
```diff
@@ -22,7 +22,7 @@ data "aws_ami" "cyhy_mongo" {
 
 resource "aws_instance" "cyhy_mongo" {
   ami = "${data.aws_ami.cyhy_mongo.id}"
-  instance_type = "${terraform.workspace == "production" ? "m4.large" : "t2.micro"}"
+  instance_type = "${terraform.workspace == "production" || terraform.workspace == "planet_piss" ? "m4.large" : "t2.micro"}"
```
dav3r, line 25: @jsf9k What do you think about using some sort of naming convention for our workspaces, e.g. "jsf9k_test" vs. "jsf9k_prod" and have our logic look for those substrings?

dav3r (commented):
@jsf9k great stuff!

terraform/cyhy_mongo_ec2.tf
jsf9k: @dav3r, that makes sense.  How about if the workspace name contains "production" then use the "real deal" instance types?

(I don't like prod because four letters seems too little...what if we hire Auguste Prodin?)

dav3r: @jsf9k, workspace name contains "production" works for me.

jsf9k: Implemented in #24.

PR #20, opened 2018-08-13
Change code to use the cheaper r4.xlarge instance type.

jsf9k:
I watched the memory consumption while running BOD scanning, and it appears to consume about 13GB.  I could probably get by with 16GB of RAM, but this is an economical option that has almost 32GB of RAM.

I did a full test run with no issues.


dav3r (approved):
This looks good to me, though I'd still like to get your thoughts on this comment: https://github.com/dhs-ncats/cyhy_amis/pull/19/files/7d511a6165ecd298e351bfa2f607a2161c68c868#r209778046

PR #21, opened 2018-08-14
Feature/egress pub

felddy:
Putting a fork in this for now.  Will probably need some additional sauce when we automate it.


jsf9k (approved):
Looks good to me.

PR #22, opened 2018-08-14
Feature/fix nessus only setup

dav3r:


felddy (approved):

jsf9k (approved):
I think this is good.  We've been using this code in production for several days now.

PR #23, opened 2018-08-14
Feature/route 53 update

felddy:
Applies all updates to Terraform that we made by hand over the last week.  This gets the rules https site working, as well as our mail sending policies.


jsf9k (approved):
I think this is good, based on my incomplete understanding of our DNS setup.

PR #24, opened 2018-08-14
Improved workspace filtering

jsf9k:
* Change workspace-dependent logic to look for the substring "production"
* Use volume tags


dav3r (approved):
Pobody's nerfect, but these changes are!

felddy (approved):

PR #27, opened 2018-08-16
Improve networking

jsf9k:
These changes simplify the networking rules and allow the scanners to be provisioned by Ansible using the bastion host as an ssh proxy.


KyleEvers (approved):
Looks good

PR #30, opened 2018-08-19
The Nessus host is not systemd

jsf9k:
It instead requires a SysV-style init file

Also a few random fixes.  I was installing `cyhy-runner` and `cyhy-commander` incorrectly, for example.


felddy (approved):
Looks good.  Thanks for wrestling the different system init scripts to the ground.

PR #32, opened 2018-08-20
Disallow the manual Nessus instances from being deleted

jsf9k:
Also do not delete the root disk upon deletion.  That way if we do mistakenly delete a manual Nessus instance we can at least retrieve the data that was stored on it.


dav3r (approved):
All good

PR #33, opened 2018-08-20
Feature/internal dns

felddy:
Private DNS is humming.
Expect some waiting while your zones are created for the first time.  
TODO: figure out how to scale ansible provisioning to resources that contain a count clause.
TODO: promote / rename locals to variables where applicable.

### DIFF
--- a/terraform/cyhy_bastion_ec2.tf
+++ b/terraform/cyhy_bastion_ec2.tf
@@ -7,6 +7,7 @@ resource "aws_instance" "cyhy_bastion" {
 
   # This is the public subnet
   subnet_id = "${aws_subnet.cyhy_scanner_subnet.id}"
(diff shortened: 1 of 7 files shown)


dav3r (approved):
Despite your best efforts, you did it!

jsf9k (commented):
If two of us happen to deploy two separate workspaces to the same region, won't there be a conflict?  Should we call these hosts `commander.<workspace_name>.local`?

Also, what about creating public DNS names for the public IPs for instances that have them?  Maybe something like `cyhy_bastion.<workspace_name>.ncats.cyber.dhs.gov`?  I know I would find that helpful, since I hate having to go to the console to look up the public IPs of the bastions.

felddy:
There won't be a conflict.  The domains and subnets are only visible within a single VPC.   I like the idea of creating the public DNS names.  I'll get on that tomorrow.   Working on some DHCP-fu right now.

PR #35, opened 2018-08-21
Improvement/dhcp options

felddy:
Configures the VPC DHCP server to push out the private domain (`local`) as the search path.  
i.e.; you can do `ssh commander` from the bastion and it will work.


dav3r (approved):
This is slick!

PR #37, opened 2018-08-21
add public DNS entry for the bastion host

felddy:
Add a spiffy record so you can find your bastion.  Also, lets introduce the world to our unique and fun workspace names!  

NOTE:  until we get _NoneNet_ to point to the new SOA we're not actually going to be able to use these records.


dav3r (approved):
Sweet!

PR #39, opened 2018-08-21
add prevent_destroy to very important zones

felddy:
Deleting these means we need to coordinate changes with OneNet, and nobody wants that.


jsf9k (approved):
Good eye! :baseball:

PR #40, opened 2018-08-21
Add support to query all regions for public IPs instead of just one

felddy:
refactor the code.   It is large enough now to need functions.
add support to gather and loop through all ec2 regions
add better comments and output for users and logs


jsf9k (approved):
I support the breaking up of this code into functions.

PR #41, opened 2018-08-21
Allow the bastion to reach the mongo host via the mongo port

jsf9k:
As requested by @dav3r.


felddy (approved):
Attitude and effort are commendable.
A pleasure to have in class.
✔︎➕

PR #43, opened 2018-08-21
add new role to configure login banners

felddy:
Things that government computers say for $1000.


jsf9k (approved):
It's a banner year for banners.

PR #44, opened 2018-08-21
Feature/create places collection

dav3r:


felddy (approved):
@dav3r what is the state of the scripts you are importing from the private repo?  Is there any special sauce there worth protecting?  I see a lot of hoops that you're jumping through to get those scripts here.  I wonder if they can't just live in the "files" of the cyhy_commander ansible role and be copied over.  Eventually we'd want this so other people can use it.  Maybe now is the time to move them and simplify this?

ansible/roles/cyhy_commander/tasks/main.yml
```diff
@@ -98,3 +98,91 @@
...
+    path: /tmp/cyhy-places/scripts
+    state: directory
+
+- name: Create the /tmp/cyhy-places/extras directory
+  file:
+    path: /tmp/cyhy-places/extras
+    state: directory
+
+- name: Download load_places.sh from GitHub
+  shell: "curl -o /tmp/cyhy-places/scripts/load_places.sh https://{{ github_oauth_token }}@raw.githubusercontent.com/jsf9k/cyhy-core/develop/var/load_places.sh"
```
jsf9k: You could use the Ansible module [`get_url`](https://docs.ansible.com/ansible/latest/modules/get_url_module.html) here and for the two similar lines below.  I think that's a little cleaner than `shell`.

ansible/roles/cyhy_commander/tasks/main.yml
```diff
@@ -98,3 +98,91 @@
...
+
+- name: Download GNIS_data_import.py from GitHub
+  shell: "curl -o /tmp/cyhy-places/scripts/GNIS_data_import.py https://{{ github_oauth_token }}@raw.githubusercontent.com/jsf9k/cyhy-core/develop/var/GNIS_data_import.py"
```
jsf9k: See comment above.

ansible/roles/cyhy_commander/tasks/main.yml
```diff
@@ -98,3 +98,91 @@
...
+    path: /tmp/cyhy-places/scripts/load_places.sh
+    mode: 0755
+
+- name: Make GNIS_data_import.py script executable
+  file:
+    path: /tmp/cyhy-places/scripts/GNIS_data_import.py
+    mode: 0755
+
+- name: Download ADDL_CYHY_PLACES.txt from GitHub
+  shell: "curl -o /tmp/cyhy-places/extras/ADDL_CYHY_PLACES.txt https://{{ github_oauth_token }}@raw.githubusercontent.com/jsf9k/cyhy-core/develop/extras/ADDL_CYHY_PLACES.txt"
```
jsf9k: See comment above.

jsf9k (approved):
This looks good to me.

jsf9k:
I'll go ahead and change those `shell` commands to use `get_url` instead.  If I don't get it working by the time I stop for the night I'll go ahead and merge this.

dav3r:
@felddy I agree that these files will be useful to others, but I think we should hold off on that until we split this repo into two (creating the AMIs vs. everything that happens after that).

@jsf9k - nice work; I see you didn't completely ignore the GitHub docs like I did. 👍

jsf9k:
I think this is in very good shape now.  If others agree with the changes I made, please go ahead and merge.

jsf9k:
@dav3r 

"If I have seen further it is by standing on the shoulders of giants."  --Sir Issac Newton

PR #45, opened 2018-08-22
Use a more customized mongo config file, similar to our previous Prod…

dav3r:
…uction config

Ensure that the mongo keyfile is present before starting mongo service


jsf9k (approved):
Assuming your intent was to enable authentication between replica sets, this looks good to me!

dav3r:
Not just authentication between replica sets, this also enables authentication to the primary mongo database.  We thought it was enabled before, but it was not - see https://docs.mongodb.com/manual/reference/configuration-options/#security.authorization

PR #46, opened 2018-08-23
Bugfix/make mongo great again

dav3r:
Turns out, it's a good idea to enable database authentication AFTER we set up the user accounts, not before. 🚫


packer/ansible/roles/mongo/tasks/main.yml
```diff
@@ -28,6 +28,15 @@
     state: present
     line: After=network.target cloud-config.service
 
+# Mongo docs recommend disabling --quiet flag in Production
+# https://docs.mongodb.com/manual/reference/configuration-options/#systemLog.quiet
+- name: Remove --quiet mongo command line option
+  lineinfile:
+    dest: /lib/systemd/system/mongod.service
+    regexp: ^ExecStart=/usr/bin/mongod --quiet --config /etc/mongod.conf
```
felddy: The regexp doesn't need to be this verbose.  Unless you're worried about not getting the quiet.

felddy (approved):

packer/ansible/roles/mongo/tasks/main.yml
dav3r: Good point- I can trim it down, but I did want the --quiet in there.  I'll get rid of the --config part.

jsf9k (approved):

packer/ansible/roles/mongo/tasks/main.yml
```diff
@@ -33,7 +33,7 @@
 - name: Remove --quiet mongo command line option
   lineinfile:
     dest: /lib/systemd/system/mongod.service
-    regexp: ^ExecStart=/usr/bin/mongod --quiet --config /etc/mongod.conf
+    regexp: ^ExecStart=/usr/bin/mongod --quiet
```
felddy: I mean, it could even be shorter.  
^ExecStart=  
There is only one ExecStart line I'm guessing, and you want it.

jsf9k:
Huzzah!

PR #50, opened 2018-08-24
Automate setup of Nessus hosts

### LINKED ISSUE #49: Nessus instance setup
@dav3r added some instructions [here](https://github.com/dhs-ncats/cyhy_amis/wiki/Nessus-Instance-Setup-Instructions) for setting up the Nessus instance.  These instructions need to be automated via Ansible.

jsf9k:
This pull request resolves #49.


ansible/roles/nessus/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+#
+# Check if a license key is registered and if not register one
+#
+- name: Check if a license key is registered
+  command: /opt/nessus/sbin/nessuscli fetch --check
+  register: registered
+  failed_when: registered.rc not in [0, 1]
+
+# - name: Register a license key if necessary
+#   command: /opt/nessus/sbin/nessuscli fetch --register-only <activationcode>
```
jsf9k: It is unclear to me how we want to store license keys and distribute them.  Or perhaps we just want to let this continue to be a manual step, which would require the updating of plugins to be manual as well.  We could add a script to the AMI that handles both, given the license key as input.

ansible/roles/nessus/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+  when: nessus.username not in users.stdout
+
+- name: Delete local copy of Nessus credentials
+  local_action:
+    module: file
+    path: /tmp/nessus_scanner_user.yml
+    state: absent
+  become: no
+
+# - name: Update plugins
```
jsf9k: I can't do this until a license key is registered.  See above.

packer/ansible/roles/docker/README.md
```diff
@@ -1,5 +1,5 @@
 docker
-=========
+======
```
felddy, line 2: It's attention to detail like this that keeps us on the CRITICAL PATH!

jsf9k (commented):
There is still some work to do here.  I need some input from the rest of the team.

packer/ansible/roles/docker/README.md
jsf9k: # :boom: CRITICAL PATH :volcano: #

felddy (approved):
Looks good.

jsf9k:
Thanks for the approval @felddy, but I intentionally labeled this PR as "work in progress" AKA "not ready to be merged."  I need some input from the rest of the team.  See my comments above.

felddy:
JUST SEND IT!

jsf9k:
You are mad, @felddy.  Simply mad!
:worried: :hospital:

ansible/roles/nessus/tasks/main.yml
felddy: Since license keys are one and done, they're not too sensitive.  I'd think a sidecar yml file would be fine.  I used to associate them to the hosts in the old ansible inventory file.  We could define a variable in terraform, not set a default, and it will prompt the user to input it.

jsf9k: Not a bad idea.  We could define said variable in the `production.tfvars` file, so the user isn't prompted in that case.

jsf9k: This change has been made.

dav3r (approved):
Fine effort @jsf9k - well done!

PR #54, opened 2018-08-24
Only update plugins and rebuild database when previously unregistered

jsf9k:
Only update the Nessus plugins and rebuild the Nessus database when Nessus does not already have an activation code.


dav3r (approved):
What a timesaver!

PR #57, opened 2018-08-25
Ramp up number of jobs per nmap and nessus host

dav3r:
I already applied this change in Production.  Checking it in to keep the team in sync.


felddy (approved):
🤘Kickstart my Heart!  🎸

PR #61, opened 2018-08-27
Double number of jobs per nmap and nessus host again

dav3r:
This change has already been applied in Production


jsf9k (approved):
Just send it!

PR #62, opened 2018-08-27
Add an Ansible role for expanding the ephemeral port range

jsf9k:
This is for the scanners, which for best results will want the full 1024-65535 range of ephemeral ports.


dav3r (approved):
Phenomephemeral work!

PR #65, opened 2018-08-29
Add flow logs that can be turned on or off via a variable

jsf9k:
* Add a variable that controls whether or not flow logs are created
* Add Terraform code to create flow logs


dav3r (approved):
Looks good to me

PR #66, opened 2018-08-29
Various changes for Production based on our testing

dav3r:
* Change to 16 jobs-per-nmap-host and 8 jobs-per-nessus-host for further Production testing
* Switch to smaller nmap EC2 instance type (t2.large -> t2.medium)
* Modify scanner security group to allow all traffic on non-SSH/Nessus ports to try to avoid connection table saturation (see https://aerissecure.com/blog/vulnerability-scanning-from-aws/)

### DIFF
--- a/terraform/cyhy_nmap_ec2.tf
+++ b/terraform/cyhy_nmap_ec2.tf
@@ -27,6 +27,9 @@ resource "aws_instance" "cyhy_nmap" {
 
   # ebs_optimized = true
   availability_zone = "${var.aws_region}${var.aws_availability_zone}"
+  # We may want to spread instances across all availability zones, however
+  # this will also require creating a scanner subnet in each availability zone
+  # availability_zone = "${element(data.aws_availability_zones.all.names, count.index)}"
 
   subnet_id = "${aws_subnet.cyhy_scanner_subnet.id}"
   private_ip = "${cidrhost(aws_subnet.cyhy_scanner_subnet.cidr_block, count.index + local.first_port_scanner)}"
@@ -62,6 +65,7 @@ resource "aws_instance" "cyhy_nmap" {
 resource "aws_ebs_volume" "cyhy_runner_data" {
   count = "${local.nmap_instance_count}"
   availability_zone = "${var.aws_region}${var.aws_availability_zone}"
+  # availability_zone = "${element(data.aws_availability_zones.all.names, count.index)}"
   type = "gp2"
   size = "${local.production_workspace ? 2 : 1}"
   encrypted = true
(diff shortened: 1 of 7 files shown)


dav3r:
Nobody approve this yet- this is a Work in Progress.

terraform/cyhy_scanner_security_group_rules.tf
```diff
@@ -21,6 +21,44 @@ resource "aws_security_group_rule" "scanner_ingress_from_private_sg_via_ssh" {
   to_port = 22
 }
 
+# Allow ingress from anywhere via all other tcp ports
```
felddy, line 24: I might add a comment in here about why we are doing this.  Describe the stateful behavior of security groups, and how it can get bogged down with connection state monitoring.  Maybe even put in the handy URL to the article.  (This will inevitably make that article go 404) 

I think these comments will be helpful for future "us", and external developers that use CyHy in the future.

dav3r: Good call- I will put something in there now and we can add to it after we (hopefully) get some further insights from AWS Support.

dav3r:
Ready for review.  It's time to close out this branch for now.

terraform/cyhy_nmap_ec2.tf
```diff
@@ -113,27 +118,401 @@ module "cyhy_nmap_ansible_provisioner_0" {
...
-#   ]
-#   envs = [
-#     "host=${aws_instance.cyhy_nmap.1.private_ip}",
-#     "bastion_host=${aws_instance.cyhy_bastion.public_ip}",
-#     "host_groups=cyhy_runner"
-#   ]
-#   playbook = "../ansible/playbook.yml"
-#   dry_run = false
-# }
+module "cyhy_nmap_ansible_provisioner_1" {
```
felddy, line 129: Maybe it's time to think about a pre-processor for the provisioners.  Getting too 💦WET💦.  Need to 🍸DRY🌵 out.

terraform/cyhy_scanner_security_group_rules.tf
```diff
@@ -21,6 +21,58 @@ resource "aws_security_group_rule" "scanner_ingress_from_private_sg_via_ssh" {
...
+#
+# IMPORTANT NOTE: The reason we allow this ingress is to avoid connection
```
felddy, line 26: Very good comment.  I am now smarter for having read it.  🎓

felddy (approved):
Looks good!  Approved despite me clicking on all the wrong buttons.

PR #67, opened 2018-08-30
Import base nessus policy

KyleEvers:


dav3r (dismissed):
I think `import_base_policy()` should be broken up into two functions:
- `upload_file()`
- `import_policy()`

Call each function in `main()`, followed by an appropriate `assert` statement to confirm that the results are as expected (similar to the pattern in `nessus6.py`).  That way, you can (somewhat) eliminate the special case you have up in `__make_request()` for `if target == POLICY_UPLOAD`.  You'll have to move some of that logic into the `elif method == 'POST'` conditional, but it will be cleaner than now.

ansible/roles/nessus/files/nessus_base.py
```diff
@@ -0,0 +1,177 @@
...
+import copy
+import glob
+import time
+import logging
+import ssl
+import json
+import requests
+import copy
+import os
+
```
jsf9k: I personally like imports to be in alphabetical order, so I can quickly find the one I'm looking for or determine that one is missing.

jsf9k (commented):
I don't have any other comments, other than those that @dav3r mentioned in his review.

ansible/roles/nessus/files/nessus_base.py
```diff
@@ -80,19 +80,24 @@ def __make_request(self, target, method, payload=None):
...
-                headers = {}
-                headers['X-Cookie'] = 'token={!s}'.format(self.token)
-                files = {'Filedata': (BASE_POLICY_FILE_NAME, open(BASE_POLICY_FILE_NAME, 'rb'), 'text/xml')}
-                response = requests.post(self.url + target, headers=headers, files=files, verify=VERIFY_SSL)
-                LOGGER.info('Policy Uploaded to Nessus Server')
-                resp_dict = json.loads(response.text)
-                response = requests.post(self.url + POLICY_IMPORT, headers=headers, data={'file':resp_dict['fileuploaded']}, verify=VERIFY_SSL)
-                LOGGER.info('Base Policy Imported to Nessus Server Policies')
-            elif method == 'GET':
+            # if target == FILE_UPLOAD:
```
jsf9k: Remove this stuff if it's not needed.  There's no need to leave dead code around that can be retrieved from the git history.

ansible/roles/nessus/files/nessus_base.py
jsf9k: If only OCD were the extent of my problems, @KyleEvers!

ansible/roles/nessus/files/nessus_base.py
```diff
@@ -83,7 +83,8 @@ def __make_request(self, target, method, payload=None, files=None):
                 response = requests.get(self.url + target, headers=headers, params=payload, verify=VERIFY_SSL)
             elif method == 'POST':
                 if files:
-                    headers = ['X-Cookie'] = 'token={!s}'.format(self.token)
+                    # This reassigning of headers is to remove the content type assignment in line 71
```
dav3r: So you think line 71 is always going to be line 71, huh?  That's adorable.

KyleEvers: No good deed goes unpunished!

ansible/roles/nessus/files/nessus_base.py
```diff
@@ -84,8 +84,7 @@ def __make_request(self, target, method, payload=None, files=None):
                 response = requests.get(self.url + target, headers=headers, params=payload, verify=VERIFY_SSL)
             elif method == 'POST':
                 if files:
-                    headers = {}
-                    headers['X-Cookie'] = 'token={!s}'.format(self.token)
+                    headers = ['X-Cookie'] = 'token={!s}'.format(self.token)
```
dav3r: I don't think this line is valid.  Shouldn't it be:
`headers = {'X-Cookie': 'token={!s}'.format(self.token)}`

dav3r (approved):
Looks great now, good work @KyleEvers

PR #68, opened 2018-08-31
Add instance for CyHy reporting

jsf9k:
I've been testing in production (where there is actual data) with `--no-log --no-dock --no-snapshots --no-pause` and this seems to be working pretty well.  I was able to generate a full set of CyHy reports.

### DIFF
--- a/ansible/roles/cyhy_reporter/tasks/main.yml
+++ b/ansible/roles/cyhy_reporter/tasks/main.yml
@@ -53,3 +53,21 @@
     path: /tmp/mongo_users.yml
     state: absent
   become: no
+
+#
+# Add users to the cyhy group
+#
+- name: Load users from YML file
+  local_action:
+    module: include_vars
+    file: ../../../../terraform/scripts/user_ssh_setup.yml
+    name: ssh_users
+  become: no
+
+- name: Add known users to the cyhy group
+  user:
+    name: "{{ item }}"
+    groups:
+      - cyhy
+    append: yes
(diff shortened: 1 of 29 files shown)


jsf9k:
Note that the `c5.2xlarge` instance type won't work yet.  I'm waiting to hear back on a limit increase request from AWS.  Just manually change it to `c5.xlarge` for now.

ansible/roles/cyhy_reporter/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+  copy:
+    dest: /etc/cyhy/cyhy.conf
+    owner: cyhy
+    group: cyhy
+    mode: 0660
+    content: |
+      [DEFAULT]
+      default-section = production
+      database-uri = mongodb://localhost:27017/
+      report-key =
```
dav3r: We need to add another S3 secret for the default report-key.  Let me know if you need any clarification on this.

terraform/cyhy_reporter_ec2.tf
```diff
@@ -34,7 +34,7 @@ resource "aws_instance" "cyhy_reporter" {
 
   root_block_device {
     volume_type = "gp2"
-    volume_size = "${local.production_workspace ? 200 : 8}"
+    volume_size = "${local.production_workspace ? 100 : 8}"
```
dav3r: We have to discuss if/where we are going to store the reports each week after they are generated and sent out.

dav3r (approved):

PR #69, opened 2018-08-31
Improvement/persist active nmap instance scans

dav3r:


jsf9k (approved):
Looks good to me :us:

PR #70, opened 2018-08-31
Grab the master report password from S3 and add it to cyhy.conf

jsf9k:
I also installed `texlive-science` in the `cyhy-reports` AMI, since it contains a font that is used in the scorecard.


dav3r (approved):
Looks great to me

PR #78, opened 2018-09-07
Install ncats-webd alongside cyhy-reports

jsf9k:
This is so the `cyhy-reports` code can directly leverage the CSV generation code in `ncats-webd` instead of reaching out to the REST interface.


dav3r (approved):
Magnifico

PR #79, opened 2018-09-07
Build nmap from latest source

jsf9k:
We're running this in production, so it may as well get merged.


dav3r (approved):
LGTM - we just have to be aware of when we are rolling out an updated version of nmap.

PR #84, opened 2018-09-10
Add cyhy_ops user to the bastion

jsf9k:
* Add an Ansible role for creating the cyhy_ops user
* Add an Ansible provisioner to the CyHy bastion EC2 instance
* Add a section to the Ansible playbook for the CyHy bastion


dav3r (approved):
Looks great!

PR #85, opened 2018-09-11
Operations/add production scanners

dav3r:


KyleEvers (approved):
It doesn't look good. But approved.

PR #86, opened 2018-09-12
Improvement/dry provisioning

felddy:
Dynamic provisioning in the house. This works around this issue: https://github.com/hashicorp/terraform/issues/953


dav3r (dismissed):
This all looks good to me.  I tested it in my local workspace and it seems to work as advertised, but I didn't try it in Production yet.

dav3r:
There is a bug, but it's not in your code.  However, this seems like a good place to fix it.

The problem is in these four places:
- https://github.com/dhs-ncats/cyhy_amis/blob/develop/terraform/cyhy_nmap_ec2.tf#L94
- https://github.com/dhs-ncats/cyhy_amis/blob/develop/terraform/cyhy_nmap_ec2.tf#L101
- https://github.com/dhs-ncats/cyhy_amis/blob/develop/terraform/cyhy_nessus_ec2.tf#L94
- https://github.com/dhs-ncats/cyhy_amis/blob/develop/terraform/cyhy_nessus_ec2.tf#L101

In all of the above, the code references `${aws_instance.cyhy_nmap.id}` or `${aws_instance.cyhy_nessus.id}`, which is the **first** nmap/nessus instance.  What we want to do is reference the specific instance that matches the count variable:
- `${aws_instance.cyhy_nmap.*.id[count.index]}`
- `${aws_instance.cyhy_nessus.*.id[count.index]}`

Unfortunately, when I make that change, I run into this open Terraform issue:
https://github.com/hashicorp/terraform/issues/14536

Example:  In my local workspace, I have zero running nmap instances and I want to ramp up to three of them.  The existing code has no problem with this because every `nmap_cyhy_runner_data_attachment` references (incorrectly) the first nmap instance (`${aws_instance.cyhy_nmap.id}`).  The corrected code however fails with an error like this:
```
* aws_volume_attachment.nmap_cyhy_runner_data_attachment: 1 error(s) occurred:

* aws_volume_attachment.nmap_cyhy_runner_data_attachment[2]: index 2 out of range for list aws_instance.cyhy_nmap.*.id (max 2) in:

${aws_instance.cyhy_nmap.*.id[count.index]}
```

Anybody have any ideas on how to fix this?

jsf9k:
The bug report you referenced says:
> Replacing array syntax with element() function (as shown in comments in code) gives the desired behaviour.

Did you try that?

dav3r:
I missed that- trying it now.

terraform/configure.py
```diff
@@ -0,0 +1,127 @@
...
+        f.write('locals {\n'.encode('utf-8'))
+        for key, variable_name in LOCAL_DEFS.items():
+            value = config.get(key)
+            f.write('    {} = {}\n'.format(variable_name, value).encode('utf-8'))
+        f.write('}\n'.encode('utf-8'))
+
+def main():
+    # get workspace
+    workspace = get_terraform_workspace()
+    print('Current Terraform workspace = ', workspace)
```
jsf9k, line 95: What about using `logging` instead of `print` statements?

dav3r:
That `element()` workaround worked- thanks @jsf9k. I had to make one other similar change to `nmap.template` and `nessus.template` get it to work.  I'm going to do a little more testing, then commit the changes.

terraform/configure.py
```diff
@@ -0,0 +1,127 @@
...
+pasting in order to get provisioners and other modules to execute correctlyself.
+
+This script can be retired once this limitation is removed from Terraform.
+See: https://github.com/hashicorp/terraform/issues/953
+
+To create a unique configuration for your own workspace edit the
+WORKSPACE_CONFIGS constant below.
+'''
+
+import sys
```
jsf9k: I like imports to be sorted in three groups in this order:
* Python built-ins
* External dependencies
* Local dependencies

Then, within each group, I like the imports to be sorted alphabetically.  This makes it easy to find imports as the list grows in length.

What do you think about doing that here?

terraform/configure.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env python3
+
+'''
+This script will configure the terraform environment based on the active
+workspace.
+
+Due to a terraform limitation, modules can not be scaled with the "count"
+keyword the same way resource can.  This leads to a great deal of copying and
+pasting in order to get provisioners and other modules to execute correctlyself.
```
jsf9k: Remove `correctlyself` and replace with `correctly`. :smiley:

jsf9k (approved):
Please consider my comments before merging.

dav3r:
@jsf9k , the `element()` workaround appears to have a problem.  

Example: When I attempt to scale up from 1 to 2 nmap instances, Terraform wants to recreate `nmap_cyhy_runner_data_attachment[0]`, which would trigger the destruction of the `cyhy_nmap[0]` instance (since that is how we safely destroy the volume attachment):
```
-/+ aws_volume_attachment.nmap_cyhy_runner_data_attachment[0] (new resource required)
      id:                                        "vai-3307678400" => <computed> (forces new resource)
      device_name:                               "/dev/xvdb" => "/dev/xvdb"
      instance_id:                               "i-06916a3ed89b85a52" => "${element(aws_instance.cyhy_nmap.*.id, count.index)}" (forces new resource)
      skip_destroy:                              "true" => "true"
      volume_id:                                 "vol-02311105b958b2c31" => "vol-02311105b958b2c31"
```

It is the same issue as mentioned in [this comment on Terraform issue 14536](https://github.com/hashicorp/terraform/issues/14536#issuecomment-316402456).

We need to figure out how to get Terraform to avoid destroying the existing data attachment(s) in this scenario.

dav3r:
My last commit fixes the issue I described [here](https://github.com/dhs-ncats/cyhy_amis/pull/86#issuecomment-421092642).

dav3r (approved):
This now gets my official 👍

PR #96, opened 2018-09-21
Operations/upgrade mongo

dav3r:
Upgrade mongo from 3.2.20 to 3.4.17

### DIFF
--- a/packer/ansible/roles/cyhy_commander/tasks/main.yml
+++ b/packer/ansible/roles/cyhy_commander/tasks/main.yml
@@ -2,8 +2,13 @@
 # tasks file for cyhy_commander
 
 #
-# Install unzip; it is needed by cyhy_core/var/load_places.sh
+# Install curl and unzip; it is needed by cyhy_core/var/load_places.sh
 #
+- name: Install curl
+  apt:
+    name: curl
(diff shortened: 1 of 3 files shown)


packer/ansible/roles/cyhy_commander/tasks/main.yml
```diff
@@ -2,8 +2,13 @@
 # tasks file for cyhy_commander
 
 #
-# Install unzip; it is needed by cyhy_core/var/load_places.sh
+# Install curl and unzip; it is needed by cyhy_core/var/load_places.sh
 #
+- name: Install curl
+  apt:
+    name: curl
```
jsf9k: Name can be a list, like this:
```
name:
  - curl
  - unzip
```

So you could combine both installs into a single Ansible command.

If you do make this change, can you also change "it is" to "they are" in the comment?

jsf9k (approved):
Approved, but please consider my one comment.

packer/ansible/roles/cyhy_commander/tasks/main.yml
dav3r: Good catches, thanks @jsf9k!

PR #100, opened 2018-09-23
Add DNS for the BOD VPC

jsf9k:
* Add DNS for the BOD VPC
* Change the names of some CyHy Terraform variables so they are distinct from the corresponding BOD variables.

Note that I do not use fixed IPs like the CyHy VPC does, but this still appears to work.

### DIFF
--- a/terraform/bod_private_dns.tf
+++ b/terraform/bod_private_dns.tf
@@ -44,7 +44,7 @@ resource "aws_route53_record" "bod_bastion_A" {
   type = "A"
   ttl = 300
   records = [
-    "${aws_instance.bod_bastion.public_ip}"
+    "${aws_instance.bod_bastion.private_ip}"
   ]
 }
 
--- a/terraform/bod_private_dns.tf
+++ b/terraform/bod_private_dns.tf
@@ -1,0 +1,153 @@
+resource "aws_route53_zone" "bod_private_zone" {
+  name = "${local.bod_private_domain}."
+  vpc_id = "${aws_vpc.bod_vpc.id}"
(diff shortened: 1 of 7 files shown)


terraform/bod_private_dns.tf
```diff
@@ -0,0 +1,153 @@
...
+  ]
+}
+
+resource "aws_route53_record" "bod_bastion_A" {
+  zone_id = "${aws_route53_zone.bod_private_zone.zone_id}"
+  name = "bastion.${aws_route53_zone.bod_private_zone.name}"
+  type = "A"
+  ttl = 300
+  records = [
+    "${aws_instance.bod_bastion.public_ip}"
```
jsf9k: Should this be the private IP?  I copied what was done in the CyHy VPC, but this doesn't seem correct.

dav3r: @felddy, care to comment about this?

felddy: I'm trying to find where @jsf9k  copied this from.  The private bastion entries I wrote don't reference the `public_ip`:
https://github.com/dhs-ncats/cyhy_amis/blob/bb304156750f0028b453ff6bef94bca92b9da700/terraform/cyhy_private_dns.tf#L35

Also, the entries in route53 look correct in production.

jsf9k: You are correct sir. I don't know where I got that from.

jsf9k:
@felddy, can you also comment on the fixed IP issue?  That didn't seem to be an issue for me.

dav3r (approved):
These changes all make sense and look good to me.

PR #101, opened 2018-09-24
Fix cron jobs

jsf9k:
* Modify the PATH of the cyhy user's crontab to include `/usr/local/bin`.
* Fix the timing of CyHy and BOD report generation cron jobs.
* Make sure relevant error outputs are piped into `syslog`.


dav3r (approved):
That's some good stuff right there

PR #103, opened 2018-09-24
Operations/upgrade to mongo 3.6

dav3r:


ansible/roles/mongo/templates/mongod.conf
```diff
@@ -15,12 +15,9 @@ systemLog:
 
 net:
     port: 27017
-    # bindIp: 0.0.0.0   # Comment out to bind to all interfaces
+    bindIp: 0.0.0.0   # Allow access from all IPv4 addresses
```
jsf9k, line 18: This could be dangerous if this host had a public IP.  I'm not sure we want to bother with passing in the instance's private IP and binding only to `localhost` and the private IP, but we should probably at least add a comment describing that this is safe in our environment.

We don't want to come along later and blindly reuse this in an environment where we lack the proper network controls.

jsf9k (approved):
Approved, but please consider my comment.

PR #104, opened 2018-09-25
Install nmap via package manager

jsf9k:
* Revert to installing nmap via the package manager
* Simplify htop tasks


dav3r (approved):
Looks good!

PR #109, opened 2018-09-27
Only create cron jobs in production workspaces

jsf9k:
Modify Ansible code to only create CyHy reporter and BOD docker cron jobs for production workspaces.


dav3r (approved):
🍰

PR #110, opened 2018-09-27
Add volumes for CyHy and BOD reports

jsf9k:
This pull request is goes along with jsf9k/cyhy-reports#9.

### DIFF
--- a/packer/ansible/playbook.yml
+++ b/packer/ansible/playbook.yml
@@ -68,6 +68,7 @@
   become: yes
   become_method: sudo
   roles:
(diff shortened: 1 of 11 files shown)


dav3r (approved):
This all looks solid to me.  Though I'm wondering why you switched to user_data_base64 - please enlighten me.

jsf9k:
@dav3r I thought that if I turned on gzip and base64 for userdata it would be smaller.

dav3r:
@jsf9k As long as it works, sounds good to me.

PR #111, opened 2018-09-27
Improvement/mongo commander on debian stretch

dav3r:


ansible/roles/mongo/tasks/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 # tasks file for mongo
-- name: Grab mongo users and paswords from S3
+- name: Grab mongo users and passwords from S3
```
jsf9k, line 3: Good catch!

jsf9k (approved):
This all looks great to me!

PR #112, opened 2018-09-27
Prefer package to apt where possible

jsf9k:
For testing I went through and built all the AMIs using this code.


dav3r (approved):
📦 📦 📦 Package all the things!

PR #114, opened 2018-10-01
The root filesystem for the BOD Docker instance needs to be larger

jsf9k:
The root filesystem for the BOD Docker instance needs to be large enough to hold all the data that docker creates.

I noticed this because the BOD scanning failed this weekend for this very reason.

I want to revert to a large drive for now, and I created #116 so we can go back later and reduce it to what is necessary.

### DIFF
--- a/terraform/bod_docker_ec2.tf
+++ b/terraform/bod_docker_ec2.tf
@@ -33,7 +33,7 @@ resource "aws_instance" "bod_docker" {
 
   root_block_device {
     volume_type = "gp2"
-    volume_size = 10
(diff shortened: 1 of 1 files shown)


dav3r:
Sounds good- we can approve it once you nail down the disk size.

jsf9k:
@dav3r I created #116 to go back and revisit this.  Can we go with the 200GB drive for now?  I need this change so I can move forward with deploying the BOD reporter changes that you approved this morning.

dav3r (approved):
Works for me- go for it.

PR #115, opened 2018-10-01
Allow for variable device names

jsf9k:
Certain AWS instance types present EBS volumes as NVMe devices, which means that the device name for the EBS volume can change depending on the underlying AWS instance type.  This was biting us with the CyHy reporter instance, since in production is uses a c5 instance type.

This code handles this by using the device name `/dev/xvdb` if we are not in a production workspace and `/dev/nvme1n1` if we are.


dav3r (approved):
Makes sense to me

PR #118, opened 2018-10-04
No more static IPs for the CyHy VPC

jsf9k:
I deployed this to the `jsf9k` workspace and it appears to be working fine.  I particularly want @felddy to review this pull request in case there is a good reason to stick with static IPs that I'm missing.

### DIFF
--- a/terraform/cyhy_private_dns.tf
+++ b/terraform/cyhy_private_dns.tf
@@ -188,6 +188,7 @@ resource "aws_route53_record" "cyhy_rev_reporter_PTR" {
 }
 
 resource "aws_route53_record" "cyhy_rev_database_PTR" {
+  count = "${local.count_database}"
   zone_id = "${aws_route53_zone.cyhy_private_zone_reverse.zone_id}"
   name    = "${format("%s.%s.%s.%s.in-addr.arpa.",
     element(split(".", aws_instance.cyhy_mongo.private_ip), 3),
@@ -197,5 +198,5 @@ resource "aws_route53_record" "cyhy_rev_database_PTR" {
   )}"
   type    = "PTR"
   ttl     = 300
-  records = [ "database1.${local.cyhy_private_domain}." ]
+  records = [ "database${count.index + 1}.${local.cyhy_private_domain}." ]
 }
--- a/terraform/cyhy_private_dns.tf
+++ b/terraform/cyhy_private_dns.tf
@@ -37,7 +37,7 @@ resource "aws_route53_record" "cyhy_bastion_A" {
   name    = "bastion.${aws_route53_zone.cyhy_private_zone.name}"
   type    = "A"
   ttl     = 300
-  records = [ "${cidrhost(aws_subnet.cyhy_scanner_subnet.cidr_block, local.the_bastion)}" ]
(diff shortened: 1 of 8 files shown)


dav3r (approved):
This all makes sense to me, but agree that @felddy should weigh in with his DNS expertise in case we are forgetting about something.

terraform/cyhy_private_dns.tf
```diff
@@ -82,9 +82,9 @@ resource "aws_route53_record" "cyhy_vulnscan_A" {
 resource "aws_route53_zone" "cyhy_scanner_zone_reverse" {
   # NOTE:  This assumes that we are using /24 blocks
   name = "${format("%s.%s.%s.in-addr.arpa.",
-    element( split(".", aws_subnet.cyhy_scanner_subnet.cidr_block) ,2),
```
felddy, line 85: Good catch... ?
How about that ") ,2)"

terraform/cyhy_private_dns.tf
```diff
@@ -118,7 +118,12 @@ resource "aws_route53_record" "cyhy_rev_3_PTR" {
 
 resource "aws_route53_record" "cyhy_rev_bastion_PTR" {
   zone_id = "${aws_route53_zone.cyhy_scanner_zone_reverse.zone_id}"
-  name    = "${local.the_bastion}.${aws_route53_zone.cyhy_scanner_zone_reverse.name}"
+  name    = "${format("%s.%s.%s.%s.in-addr.arpa.",
```
felddy, line 121: We'll have to revisit this for IPv6.

terraform/cyhy_private_dns.tf
```diff
@@ -161,17 +176,26 @@ resource "aws_route53_zone" "cyhy_private_zone_reverse" {
...
+    element(split(".", aws_instance.cyhy_mongo.private_ip) ,3),
+    element(split(".", aws_instance.cyhy_mongo.private_ip) ,2),
+    element(split(".", aws_instance.cyhy_mongo.private_ip) ,1),
+    element(split(".", aws_instance.cyhy_mongo.private_ip) ,0),
+  )}"
   type    = "PTR"
   ttl     = 300
-  records = [ "database${count.index + 1}.${local.cyhy_private_domain}." ]
+  records = [ "database1.${local.cyhy_private_domain}." ]
```
felddy: What happened to our multiple database support?

felddy (changes requested):
I think there might be an issue with hard coding in `database1` in `cyhy_rev_database_PTR`.   We still have `count_database` defined in `locals.tf`

terraform/cyhy_private_dns.tf
jsf9k: Now fixed.

terraform/cyhy_private_dns.tf
jsf9k: Agreed, but that was true before too.

terraform/cyhy_private_dns.tf
jsf9k: It's back.

felddy (approved):
Is nice.

PR #120, opened 2018-10-05
Persist journald logs across reboots

jsf9k:
* Add ansible role to persist journald storage across reboots
* Apply above ansible role to all AMIs

I'll rebuild all the AMIs once this is approved and merged.  I tested it with the bastion hosts in the `jsf9k` workspace.


dav3r (approved):
Crisp and clean 👍

PR #121, opened 2018-10-05
Don't apply journald tweak to the Nessus instance

jsf9k:
Only apply the changes that persist the `journald` logs to systems that actually run `journald`.


dav3r (approved):
Guess that's kind of important 😄

PR #122, opened 2018-10-06
Increase root volume sizes

jsf9k:
Bump up the size of some root volumes now that we are writing `journald` logs there.

The CyHy Mongo and CyHy reporter root volumes were fullish, so I decided to increase them a little bit.  The other instances appear fine for now.

I'm not against bumping *everyone* up to a 16GB or 20GB root volume, either.  **What do @felddy, @dav3r, and @KyleEvers think makes sense?**


dav3r:
If it will save us some headaches later, I'm fine with just increasing all instance root volumes to 20GB.

dav3r (approved):

PR #124, opened 2018-10-09
Add cyhy-mailer to cyhy reporter AMI

jsf9k:
* Split up the `bod_docker` role in the first (Packer) Ansible into separate `orchestrator` and `cyhy_mailer` roles
* Split up the `bod_docker` role in the second (Terraform) Ansible into separate `orchestrator` and `cyhy_mailer` roles
* Make Ansible playbook changes in accordance with the previous two steps
* Centralize the placing of users into supplementary roles (which can only be done after `cloud-init` has run) into a `groups` role that is run on all hosts.


dav3r (approved):
@jsf9k this looks great to me- nice work!

PR #126, opened 2018-10-10
Only check for 'prod' in workspace name, rather than 'production'.

dav3r:
This is so we can name our production workspaces 'prod-a' and 'prod-b'.


jsf9k (approved):
:+1:

PR #130, opened 2018-10-19
Break out cyhy-core install and retrieval of GitHub OAuth token into roles

jsf9k:
Both these tasks are repeated in the Ansible code, so it makes sense to break them out into their own roles.

### DIFF
--- a/packer/ansible/roles/cyhy_core/meta/main.yml
+++ b/packer/ansible/roles/cyhy_core/meta/main.yml
@@ -1,0 +1,57 @@
+galaxy_info:
+  author: your name
+  description: your description
+  company: your company (optional)
+
+  # If the issue tracker for your role is not on github, uncomment the
+  # next line and provide a value
+  # issue_tracker_url: http://example.com/issue/tracker
+
+  # Some suggested licenses:
+  # - BSD (default)
+  # - MIT
+  # - GPLv2
+  # - GPLv3
+  # - Apache
+  # - CC-BY
+  license: license (GPLv2, CC-BY, etc)
+
+  min_ansible_version: 1.2
+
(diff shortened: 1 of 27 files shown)


packer/ansible/roles/cyhy_core/meta/main.yml
```diff
@@ -0,0 +1,57 @@
...
+
+  galaxy_tags: []
+    # List tags for your role here, one per line. A tag is a keyword that describes
+    # and categorizes the role. Users find roles by searching for tags. Be sure to
+    # remove the '[]' above, if you add tags to this list.
+    #
+    # NOTE: A tag is limited to a single word comprised of alphanumeric characters.
+    #       Maximum 20 tags per role.
+
+dependencies: []
```
dav3r: @jsf9k Should we include github_oauth as a role dependency here?
https://docs.ansible.com/ansible/2.5/user_guide/playbooks_reuse_roles.html#role-dependencies

dav3r (approved):
Looks solid @jsf9k - I only have that one comment about dependencies.

packer/ansible/roles/cyhy_core/meta/main.yml
jsf9k: Thanks for pointing that out @dav3r!  I will go ahead and add `github_oauth` as a dependency where appropriate, and I will create another PR where I take full advantage of dependencies to simplify our Ansible code.

jsf9k:
@dav3r, please take another look.  If you're still happy then I'll merge this pull request and start applying dependencies globally.

dav3r (approved):
Looks great @jsf9k - nice work!

PR #131, opened 2018-10-23
Make more use of ansible dependencies

jsf9k:
In this pull request I make more use of Ansible dependencies in the Ansible code that is run by `packer`.  If folks like this pull request then I will make an analogous pull request for the Ansible code that is run by `terraform`.


dav3r (approved):
This is the 🐝🐝🐝 knees - so much cleaner than before!  I'm all for doing this in our terraform ansible blocks as well.

PR #132, opened 2018-10-23
List chown_cyhy_dirs as a dependency for several roles

jsf9k:


dav3r (approved):
👍 Looks good.  Nice catches on those other minor corrections you made too.

jsf9k:
Actually, I'm not sure the `chown_cyhy_dirs` bit is correct.  It's not a dependency since it really needs to run _after_ `cyhy-reports`, for example, is installed.  (I'm realizing this as I watch it apply to the `jsf9k` workspace.)

jsf9k:
Nevermind, I think everything is OK.  The point of the `chown_cyhy_dirs` is to `chown` to `cyhy:cyhy` all the cyhy-stuff that was installed in the `packer` Ansible.  (We couldn't do it then because the `cyhy` user didn't exist yet.)

All the configuration that is done in the `cyhy_*` roles in the `terraform` Ansible is done with the intended user and group.

Do you agree, @dav3r?

dav3r:
@jsf9k - Yes, that makes sense.

PR #133, opened 2018-10-24
Move some stuff around between cyhy-feeds Ansibles

jsf9k:
I didn't like that scripts were being installed in the Terraform Ansible, so I fixed that.  I took the opportunity to reduce the amount of code using Ansible `loop`s in several places.

I also ran into [this problem with `gpg2`](https://discuss.circleci.com/t/error-sending-to-agent-inappropriate-ioctl-for-device/17465), perhaps because I have `gpg-agent` running locally.  In any event, adding the `--batch` option to the `gpg2` commands gets rid of the issue and otherwise changes nothing.

### DIFF
--- a/ansible/roles/cyhy_feeds/meta/main.yml
+++ b/ansible/roles/cyhy_feeds/meta/main.yml
@@ -53,4 +53,4 @@ galaxy_info:
     #       Maximum 20 tags per role.
 
 dependencies:
-  - role: chown_cyhy_dirs
+  - chown_cyhy_dirs
--- a/ansible/roles/cyhy_feeds/tasks/main.yml
+++ b/ansible/roles/cyhy_feeds/tasks/main.yml
@@ -1,19 +1,76 @@
 ---
 # tasks file for cyhy_feeds
-- name: Grab GitHub OAuth token from S3
+
+#
+# Grab and copy some secrets from S3
+#
+- name: Grab secrets from S3
   local_action:
     module: aws_s3
     bucket: ncats-cyhy-secrets
-    object: github_oauth_token.yml
-    dest: /tmp/github_oauth_token.yml
+    object: "{{ item }}"
+    dest: "/tmp/{{ item }}"
     mode: get
   become: no
+  loop:
+    - cyhy-data-extract.cfg
+    - ncats_lab_public_gpg.key
+    - ncats_lab_private_gpg.key
+    - nsd_cyhy_daily_extract_public.key
+    - gpg_keys.trust
 
-- name: Load GitHub OAuth token from YML file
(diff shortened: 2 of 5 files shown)


jsf9k:
@KyleEvers tested and found that running the script fails with:
```
Traceback (most recent call last):
  File "cyhy-data-extract.py", line 31, in <module>
    from requests_aws4auth import AWS4Auth
  File "/usr/local/lib/python2.7/dist-packages/requests_aws4auth/__init__.py", line 204, in <module>
    from .aws4auth import AWS4Auth, StrictAWS4Auth, PassiveAWS4Auth
  File "/usr/local/lib/python2.7/dist-packages/requests_aws4auth/aws4auth.py", line 26, in <module>
    from requests.auth import AuthBase
  File "/usr/lib/python2.7/dist-packages/requests/__init__.py", line 63, in <module>
    from . import utils
  File "/usr/lib/python2.7/dist-packages/requests/utils.py", line 24, in <module>
    from ._internal_utils import to_native_string
  File "/usr/lib/python2.7/dist-packages/requests/_internal_utils.py", line 11, in <module>
    from .compat import is_py2, builtin_str, str
  File "/usr/lib/python2.7/dist-packages/requests/compat.py", line 46, in <module>
    from .packages.urllib3.packages.ordered_dict import OrderedDict
ImportError: No module named ordered_dict
```

The problem turns out to be that the `apt` package `python-boto` is installed in our base image, and `python-boto` in turn pulls in the package `python-requests`.  That means that when `pip` runs (via `packer`) it sees that `requests` is already installed.  We end up with versions of `requests_aws4auth` and `requests` that do not agree.

I'm testing whether simply passing `--upgrade` to `pip` is sufficient to fix this issue.

jsf9k:
@KyleEvers, it appears to work now.  Please review again when you have time.  

The script still fails, but only because I don't have a real database in my test Terraform workspace:
```
$ sudo -u cyhy python2.7 cyhy-data-extract.py --section production --aws --config cyhy-data-extract.cfg
Traceback (most recent call last):
  File "cyhy-data-extract.py", line 210, in <module>
    main()
  File "cyhy-data-extract.py", line 152, in main
    all_orgs = db.RequestDoc.get_all_descendants('ROOT')
  File "/usr/local/lib/python2.7/dist-packages/cyhy/db/database.py", line 917, in get_all_descendants
    raise ValueError(owner + ' has no request document')
ValueError: ROOT has no request document
```

I can retest in a different workspace, if you want to give me one where
- There is enough of a database to get past this error
- I can redeploy the `cyhy-feeds` instance

KyleEvers (approved):
Looks good. I like how much cleaner the loops are.

PR #135, opened 2018-10-26
Use GitHub OAuth role from GitHub

jsf9k:
Remove it from this repository.

If the other devs like this approach then we can extract a lot of other roles that are likely even more reusable than `github_oauth`.


felddy (approved):
I like these changes, and would be happy to see us adopt it across all of our projects. So dry!

jsf9k:
I'd still like @dav3r and @KyleEvers to weigh in before merging.  I too think that in most cases separate and reusable Ansible roles are the way forward, but I want others to voice concerns if they have them.

KyleEvers:
I also like these changes. Looks good.

README.md
```diff
@@ -1,32 +1,46 @@
...
+because the AMIs produced by those lines are large and need extra time
+to be copied, as discussed
 [here](https://github.com/hashicorp/packer/issues/6536#issuecomment-407925535).
 
-Build Terraform-based infrastructure with:
+Also note that
+```
+ansible-galaxy install --force -r packer/ansible/requirements.yml
+```
+will update the roles that are being pull from external sources.  This
```
dav3r:
```suggestion
will update the roles that are being pulled from external sources.  This
```

dav3r (changes requested):
I'm a big fan of this approach- let's keep it going. 👍

dav3r (approved):

PR #137, opened 2018-11-02
Modify the prevent_destroy toggle script for OSX

jsf9k:
The `-i` flag doesn't work the same in the GNU and BSD implementations of `sed`, so I have to resort to some tomfoolery to work around that.

The BSD implementation of `sed` also doesn't allow the `\s` notation I was using for whitespace matching, so I modified the regex to use `[[:space:]]` instead.

### DIFF
--- a/terraform/scripts/toggle_prevent_destroy
+++ b/terraform/scripts/toggle_prevent_destroy
@@ -2,6 +2,10 @@
 
 before=false
 after=true
+# We need the sed script because the file redirection doesn't play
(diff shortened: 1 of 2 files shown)


jsf9k:
GIve this a try, @dav3r.  Let me know if it works or not.

dav3r (approved):
@jsf9k - it worked perfectly for me.  Looks good, thanks for the fix!

jsf9k:
Totally awesome, dude! :surfing_man:

PR #138, opened 2018-11-03
Add TravisCI config

jsf9k:
This took a few iterations to get right.  One problem is that Python 3.7 (required for `terraform/configure.py`) isn't supported in TravisCI unless you specifically request a Xenial VM.  Once I did that, I realized that `packer validate` can't completely validate our Packer code because the Xenial VMs run in GCE.  (On the Trusty VMs, which run in EC2, `packer validate` works just fine.) Therefore I had to add the `-syntax-only` flag to `packer validate`, and that command isn't validating as much as it could.

I could fix the issue by adding (encrypted) AWS credentials to the TravisCI configuration, but it seems hokey to create a user just for that.

The TravisCI situation would be improved if we split the AMI generation code into its own repository, since then I could run `packer validate` in TravisCI's Trusty environment.  In one sense the complication of testing both Packer and Terraform code in the same repository is due to TravisCI's deficiencies.  To my mind, though, it is another indication that this repo contains _too much stuff_ and should be broken up.  That's my $0.02.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 language: python
 
+env:
+  global:
+    - AWS_DEFAULT_REGION=us-east-1
+    - AWS_ACCESS_KEY_ID=AAAAAAAAAAAAAAAAAAAA
+    - AWS_SECRET_ACCESS_KEY=ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ
+
 matrix:
   include:
     - python: 3.7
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,30 @@
+language: python
+
+matrix:
+  include:
+    - python: 3.7
```
dav3r, line 7: Might be worth adding a couple of comments about why we need python 3.7 and xenial in here, even though you explained it well in your PR comments.

dav3r (approved):
This looks good to me, I just made a small request for additional comments in `.travis.yml`. 👍

.travis.yml
jsf9k: Good call.

PR #140, opened 2018-11-05
Remove local cyhy_logrotate role and use the one in GitHub instead

jsf9k:


dav3r (approved):
Role with the changes!  🎹 🐟

PR #143, opened 2018-11-15
Add missing route table association

jsf9k:
There was no route table association to tie the private route table to the private subnet.  The private subnet was using the default (portscan) route table instead.

I discovered this because it wasn't possible to connect to the MongoDB instance in the CyHy VPC from the BOD VPC.


dav3r (approved):
Great catch @jsf9k

PR #145, opened 2018-11-19
Improvement/remove nat gateways

dav3r:
Since the NAT gateways did not give us the throughput that we needed, we removed them.  To keep our public IP footprint to the 4 elastic IPs that we have already publicized, we switch from many small nmap instances to one large nmap instance (until a better solution can be achieved).


jsf9k (approved):
Top notch commits by all involved.  :1st_place_medal:

PR #146, opened 2018-11-27
Improvement/use contiguous cidr block

dav3r:
Also, switch back to using 64 small portscan instances (vs. 1 huge portscan instance)

### DIFF
--- a/terraform/cyhy_nmap_ec2.tf
+++ b/terraform/cyhy_nmap_ec2.tf
@@ -35,7 +35,7 @@ resource "aws_instance" "cyhy_nmap" {
 
   root_block_device {
     volume_type = "gp2"
-    volume_size = "${local.production_workspace ? 8 : 8}"
+    volume_size = "${local.production_workspace ? 20 : 8}"
     delete_on_termination = true
   }
 
--- a/terraform/cyhy_nmap_ec2.tf
+++ b/terraform/cyhy_nmap_ec2.tf
@@ -22,7 +22,7 @@ data "aws_ami" "nmap" {
 
 resource "aws_instance" "cyhy_nmap" {
   ami = "${data.aws_ami.nmap.id}"
-  instance_type = "${local.production_workspace ? "r5.24xlarge" : "t3.micro"}"
+  instance_type = "${local.production_workspace ? "t3.micro" : "t3.micro"}"
   count = "${local.nmap_instance_count}"
 
   ebs_optimized = "${local.production_workspace}"
@@ -35,7 +35,7 @@ resource "aws_instance" "cyhy_nmap" {
 
   root_block_device {
     volume_type = "gp2"
-    volume_size = "${local.production_workspace ? 100 : 8}"
+    volume_size = "${local.production_workspace ? 8 : 8}"
     delete_on_termination = true
   }
 
@@ -55,7 +55,7 @@ resource "aws_instance" "cyhy_nmap" {
 # changes.
 data "aws_eip" "cyhy_nmap_eips" {
   count = "${local.production_workspace ? local.nmap_instance_count : 0}"
-  public_ip = "${element(var.cyhy_portscan_elastic_ips, count.index)}"
+  public_ip = "${cidrhost(var.cyhy_elastic_ip_cidr_block, var.cyhy_portscan_first_elastic_ip_offset + count.index)}"
 }
 
 # The Elastic IPs for the *non-production* CyHy nmap scanner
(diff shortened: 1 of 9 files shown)


terraform/cyhy_nmap_ec2.tf
```diff
@@ -22,7 +22,7 @@ data "aws_ami" "nmap" {
 
 resource "aws_instance" "cyhy_nmap" {
   ami = "${data.aws_ami.nmap.id}"
-  instance_type = "${local.production_workspace ? "r5.24xlarge" : "t3.micro"}"
+  instance_type = "${local.production_workspace ? "t3.micro" : "t3.micro"}"
```
jsf9k, line 25: This line and the `volume_size` line below it could obviously be simplified if the values don't depend on whether or not we're in a production workspace.  What is the intent?  Are you going to see if this works out and follow up?

terraform/cyhy_nmap_ec2.tf
```diff
@@ -35,7 +35,7 @@ resource "aws_instance" "cyhy_nmap" {
 
   root_block_device {
     volume_type = "gp2"
-    volume_size = "${local.production_workspace ? 100 : 8}"
+    volume_size = "${local.production_workspace ? 8 : 8}"
```
jsf9k: Is 8GB what we had before?  Seems small.

terraform/cyhy_nmap_ec2.tf
```diff
@@ -104,7 +104,7 @@ resource "aws_ebs_volume" "nmap_cyhy_runner_data" {
   availability_zone = "${var.aws_region}${var.aws_availability_zone}"
   # availability_zone = "${element(data.aws_availability_zones.all.names, count.index)}"
   type = "gp2"
-  size = "${local.production_workspace ? 128 : 1}"
+  size = "${local.production_workspace ? 2 : 1}"
```
jsf9k, line 107: Again, is 2GB the value we used before?

terraform/variables.tf
```diff
@@ -66,14 +66,17 @@ variable "cyhy_archive_bucket_name" {
...
-  default = []
+variable "cyhy_portscan_first_elastic_ip_offset" {
+  description = "The offset of the address (from the start of the elastic IP CIDR block) to be assigned to the *first* CyHy portscan instance.  For example, if the CIDR block is 192.168.1.0/24 and the offset is set to 10, the first portscan address used will be 192.168.1.10.  This is only used in production workspaces.  Each additional portscan instance will get the next consecutive address in the block.  NOTE: This will only work as intended when a contiguous CIDR block of EIP addresses is available."
+  default = 0
+}
+
+variable "cyhy_vulnscan_first_elastic_ip_offset" {
+  description = "The The offset of the address (from the start of the elastic IP CIDR block) to be assigned to the *first* CyHy vulnscan instance.  For example, if the CIDR block is 192.168.1.0/24 and the offset is set to 10, the first vulnscan address used will be 192.168.1.10.  This is only used in production workspaces.  Each additional vulnscan instance will get the next consecutive address in the block.  NOTE: This will only work as intended when a contiguous CIDR block of EIP addresses is available."
```
jsf9k: Duplicate "The".

jsf9k (approved):
Approved, but please review the comments I made.

terraform/cyhy_nmap_ec2.tf
dav3r: Yes, I left it like that in case we end up switching to different instance types in the future.  Plus, it's consistent with the code over in the vulnscan instance.  If it looks like we are never going to have different instance types between Production and non-Prod, then I'm all for making the change to simplify this line.

terraform/cyhy_nmap_ec2.tf
dav3r: You're right- we used to have 20 GB here before (see [here](https://github.com/dhs-ncats/cyhy_amis/commit/b0b6b4175d080ced414494fe8691f03405b41207#diff-0ec35a026f754ebbbc04103a8e4c9031)).  I will change this back to 20 shortly.

terraform/cyhy_nmap_ec2.tf
dav3r: 2GB is what [we used to have for this;](https://github.com/dhs-ncats/cyhy_amis/commit/b0b6b4175d080ced414494fe8691f03405b41207#diff-0ec35a026f754ebbbc04103a8e4c9031) the runner doesn't require that much space if we only have 12 concurrent jobs running on each portscan instance.

KyleEvers (approved):

terraform/cyhy_nmap_ec2.tf
dav3r: Resolved in [this commit](https://github.com/dhs-ncats/cyhy_amis/pull/146/commits/42a97111fa99339a441e52b207a0c0e465c6778a)

terraform/variables.tf
dav3r: Resolved in [this commit](https://github.com/dhs-ncats/cyhy_amis/pull/146/commits/bd4335618c1a62fc7456a4f1b836620b10bb933f)

jsf9k (approved):
Like a colossus he doth bestride the world!

PR #147, opened 2018-11-30
Disable the CyHy reporting cron job (for now).

jsf9k:
We can re-enable it once the BOD scanning and the CyHy reporting can share a common redis instance.  Also, the BOD scanning is changing week to week at this point, and I want the opportunity to manually verify the scan results before the CybEx scorecard is created.


KyleEvers (approved):
Perhaps open an issue for an improvement stating that we should get BOD scanning and the CyHy reporting on a shared redis instance so this does not get lost in the shuffle.

PR #151, opened 2018-12-20
Remove stray mode line

jsf9k:
A mode line makes no sense next to a shell command.  And it makes Ansible fail (at least Ansible version 2.7.4).


dav3r (approved):
That is odd- it did not fail when I ran this in my workspace yesterday.  Maybe it's an ansible version difference; I'm running ansible 2.6.3.

PR #152, opened 2019-01-02
add cd to directory of script because the script expects the output d…

KyleEvers:
…irectory to be in the same directory


dav3r (approved):
🚶

PR #153, opened 2019-01-05
Lambda functions executing in BOD VPC behind a NAT gateway

jsf9k:
In this pull request I create a separate subnet in the BOD VPC in which the BOD scanning Lambda functions can execute.  I also change the BOD Docker EC2 instance to use an instance role for permissions instead of users.

I have also moved the creation of the BOD scanning Lambda functions into this Terraform code.  Previously these functions were created as a separate manual step, but now they need to be created here because they require information about the BOD VPC.

dhs-ncats/orchestrator#53 contains changes that are required as a result of the move from AWS users to an EC2 instance role.  This pull request and that one go hand in hand.

See also dhs-ncats/dmarc-import-terraform#5.


jsf9k:
@cisagov/team-ois I retested this PR and it looks to be working.  Please review so we can get this ancient PR out the door. :older_man:

terraform/bod_vpc.tf
```diff
@@ -73,7 +86,22 @@ resource "aws_eip" "bod_eip" {
 
 # The NAT gateway for the VPC
 resource "aws_nat_gateway" "bod_nat_gw" {
-  allocation_id = "${aws_eip.bod_eip.id}"
+  # This affront to the laws of nature merits an explanation.
```
dav3r, line 89: \<sarcasm\>So happy to see this ugliness lives on...\<\/sarcasm\>  🤢

dav3r (approved):
This all looks groovy to me - nice work @jsf9k  🍪

README.md
```diff
@@ -1,5 +1,8 @@
 # NCATS AWS AMIs :dvd: #
 
+[![Build
+Status](https://travis-ci.org/cisagov/cyhy_amis.svg?branch=develop)](https://travis-ci.org/cisagov/cyhy_amis)
```
dav3r, line 4: Nice! 🏗

PR #154, opened 2019-01-08
Use boto3 for sending emails instead of connecting to SMTP directly

jsf9k:
This is the preferred way of doing things in AWS, and we were only doing it the other way because that's how we were doing it when everything was in the CAL.

This pull request also allows the CyHy reporter and BOD docker EC2 instances to send their own reports, so the CyHy reports will no longer need to be moved from the reporter instance to the docker instance.

Note that this PR builds on #153 and is dependent on dhs-ncats/cyhy-mailer#33. 
Therefore this pull request should only be merged after those two.


dav3r (commented):
Is it worth removing the cyhy-mailer_secrets from S3?

jsf9k:
I will remove those secrets once this is merged and works once in production.  I will also delete the SMTP user at that time.

terraform/bod_docker_acl_rules.tf
```diff
@@ -22,29 +22,7 @@ resource "aws_network_acl_rule" "docker_ingress_anywhere_via_ephemeral_ports_tcp
...
-  protocol = "udp"
-  rule_number = 112
-  rule_action = "allow"
-  cidr_block = "8.8.4.4/32"
-  from_port = 1024
-  to_port = 65535
-}
-
-# Allow outbound HTTP, HTTPS, and FTP anywhere
+# Allow outbound HTTP, HTTPS, SMTP (587), and FTP anywhere
```
dav3r: Do we need to remove `SMTP (587)` from this comment?

dav3r (approved):
Top notch as always @jsf9k  🎩

terraform/bod_docker_acl_rules.tf
jsf9k: You are correct, sir!  I will remove it now.

jsf9k: Removed in 7ff20c3.

PR #155, opened 2019-01-15
add trustymail, pshtt, and sslyze data to cron job

KyleEvers:


ansible/roles/cyhy_feeds/tasks/main.yml
```diff
@@ -107,9 +107,12 @@
       report-key =
 
       [production]
-      database-uri = mongodb://commander:{{ mongo_users.other_users | selectattr("user", "eq", "commander") | map(attribute="password") | first }}@database1:27017/cyhy
+      database-uri = mongodb://commander:{{ mongo_users.other_users | selectattr("user", "eq", "commander") | map(attribute="password")| first }}@database1:27017/cyhy
```
jsf9k: Why did you get rid of the space before the `|` here?

ansible/roles/cyhy_feeds/tasks/main.yml
```diff
@@ -107,9 +107,12 @@
...
       database-name = cyhy
 
+      [production_bod_scan]
+      database-uri = mongodb://scan-reader:{{ mongo_users.other_users | selectattr("user", "eq", "scan-reader") | map(attribute="password")| first }}@database1:27017/scan
```
jsf9k: Here too.

jsf9k (changes requested):
I have a minor question about a deleted space.  Once that is resolved then this looks fine to me.

jsf9k (approved):

ansible/roles/cyhy_feeds/tasks/main.yml
KyleEvers: I fixed them. Atom was rendering it weird because the line was so long and I ended up deleting the space by mistake.

jsf9k: Ah.

PR #156, opened 2019-01-17
Move to newer instance types

jsf9k:
This code migrates all AWS instances to the latest generation of instance types.  These newer instance types all support NVMe drives, so that required some changes in the way non-root drives are mounted.  The new method is much more reusable than the old method.

This update also requires rebuilding all the AMIs, since things are considerably better behaved if the AMI is built on hardware that supports NVMe.

Note that our Nessus AMI cannot be built using newer generation instance type, since the base AMI it is based on does not support NVMe.  (I tried.)

Finally, I have marked this PR as a work in progress because I have not yet modified the MongoDB instance and AMI.  That case is a little more complicated because of the order in which the non-root drives are mounted matters.  (Fear not!  The ever-despised systemd [offers a solution](https://github.com/systemd/systemd/commit/3519d230c8bafe834b2dac26ace49fcfba139823).)  I will be working on that, *but if folks think it makes sense to go ahead and merge this one and update the Mongo instance in a separate PR I'm fine with that*.


terraform/scripts/reporter_disk_setup.yml
```diff
@@ -3,17 +3,17 @@
 mounts:
   # Add new mount to /etc/fstab, but don't automatically mount it.  We
   # must handle nested mounts in bootcmd section below
-  - [ ${device}, /var/cyhy/reports/output, xfs, 'defaults,noauto', '0', '2' ]
+  - [ /dev/nvme1n1, /var/cyhy/reports/output, xfs, 'defaults,noauto', '0', '2' ]
```
dav3r: Nice to see that this got simpler now that both Production and non-Production instance types are NVMe. 👍

jsf9k:
FWIW, I have tested these changes by building my own AMIs and testing them in my own Terraform workspace.

dav3r (approved):
Everything in here looks great to me- big fan of all of these changes! 🍾 Fine work @jsf9k 
Also, I agree with your thought about saving the Mongo-related changes for a separate PR.  Maybe create an issue so we don't forget about it?

PR #159, opened 2019-01-18
Upgrade mongo instance type

jsf9k:
The mongo EC2 instance now uses latest-generation hardware that supports NVMe volumes.  This necessitated some minor changes to cloud-init scripts for other instances that have additional volumes as well.

In addition, the mongo EC2 instance as well as EC2 instances that involve cyhy-runner will now wait until all additional filesystems are mounted before starting the relevant services (`mongod` and `cyhy-runner`).

Finally, this pull request includes a minor change to ensure that `mongod` is started before `cyhy-commander`.  The lack of such a dependency was occasionally causing a failure when `cyhy-commander` started up due to the race condition.

This code has been tested via deployment to a test Terraform workspace (`jsf9k2`).

### DIFF
--- a/packer/ansible/roles/cyhy_commander/files/cyhy-commander.service
+++ b/packer/ansible/roles/cyhy_commander/files/cyhy-commander.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=cyhy-commander service
-After=network.target
+After=network.target mongod.service
 
 [Service]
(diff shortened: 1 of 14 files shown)


dav3r (approved):
Excellent @jsf9k ; this is a very welcome set of improvements! 🍔

PR #160, opened 2019-01-18
Improvement/mongo log rotation

### LINKED ISSUE #91: Rotate and compress MongoDB log
I don't think this was being done in the past, but we now only have an 8GB volume for our Mongo log, so this could become an issue soon.

Note: It seems that you have to restart the mongo service as part of the log rotation, but I'm not sure why we would need to if we use `copytruncate`.    See https://docs.mongodb.com/manual/tutorial/rotate-log-files/

dav3r:
This PR implements log rotation for mongo.  Reminder that since it uses external ansible roles, those must be installed prior to deploying this change (see the [README](https://github.com/dhs-ncats/cyhy_amis/blob/develop/README.md) for details).

This pull request resolves #91


jsf9k (approved):
This all looks good to me! :hotdog:

Issue #129, opened 2018-10-19
Create /etc/cyhy/cyhy.conf in multiple roles


KyleEvers:
We create `/etc/cyhy/cyhy.conf` in multiple roles, such as feeds and commander, which is redundant and could cause potential errors when both roles are applied to the same instance.  

See [here](https://github.com/dhs-ncats/cyhy_amis/pull/128#discussion_r225545701) for details.

dav3r:
Closed by PR #161

PR #161, opened 2019-01-23
Bugs and fixes encountered whilst switching to prod-a

jsf9k:
You will want to remove the existing `/etc/cyhy/cyhy.conf` file on `database1.workspace.cyhy` before applying these changes.  This is because the changes to the `cyhy-commander` and `cyhy-feeds` Ansible code use the `blockinfile` module.  This module relies on the presence or absence of guard comments to know whether a given block has been added already or not.  Therefore if there is an existing `/etc/cyhy/cyhy.conf` file with the sections but not the guard comments it will result in duplicate sections.


ansible/roles/cyhy_feeds/tasks/main.yml
```diff
@@ -113,6 +118,17 @@
...
+
+- name: Check if desired additions are in cyhy.conf and if not append them
+  lineinfile:
+    dest: /etc/cyhy/cyhy.conf
+    line: '{{ item }}'
+  with_items:
+    - '[production_bod_scan]'
+    - database-uri = mongodb://scan-reader:{{ mongo_users.other_users | selectattr("user", "eq", "scan-reader") | map(attribute="password")| first }}@database1:27017/scan
+    - 'database-name = scan'
+
```
jsf9k, line 156: I don't think this is a good way to handle this problem.  Why include the `production` section from the commander?

Better might be to create the `cyhy.conf` file that feeds needs if one does not already exist.  If one does exist then add the desired lines as you are doing.  You will want to add a  `when: cyhy_conf_result.stat.exists == True` to this block since there's no need for it to run if we're creating the `cyhy.conf` file from whole cloth.  Then there should be a parallel structure in the `cyhy_commander` Ansible code.

That way the correct file will get created regardless and feeds need not know anything about what commander wants in its `cyhy.conf` file.

jsf9k (commented):
I'm not a fan of the current fix for the feeds-commander `cyhy.conf` conflict.  I will try to come back to it tonight with a better solution.

ansible/roles/cyhy_feeds/tasks/main.yml
jsf9k: OK, I think I have a better solution in ee1697b.  Note that the feeds and commander Ansible roles now have a parallel structure.  It should not matter which one is run first, nor should it matter if they are run multiple times.

@dav3r and @KyleEvers, please let me know what you think.  If you're happy with my change then I think this pull request is ready to be merged.  (I would approve it, but I can't approve a pull request that I created myself.)

jsf9k: It seems I misunderstood what was going on.  feeds needs the commander/production `cyhy.conf` section.  This is corrected in commit 0dffd4a.

jsf9k:
@dav3r and @KyleEvers, please have a look at my latest changes.  This is a semi-temporary solution. 

The real solution is to create a separate Ansible role that creates a `cyhy.conf` with credentials for each and every user in `mongo_users.other_users`.  That way there will be a conf file section available for any Mongo user that anyone wants to use.  This will cause some heartburn, though, since the credential sections will have to be named using the MongoDB username.  (Right now they are named willy-nilly.)  That will necessitate some changes elsewhere in the code where named sections from `cyhy.conf` are explicitly referenced.

ansible/roles/cyhy_feeds/tasks/main.yml
```diff
@@ -111,17 +120,33 @@
       database-uri = mongodb://localhost:27017/
       report-key =
 
-      ; BEGIN ANSIBLE MANAGED BLOCK CYHY FEEDS
+      ; BEGIN ANSIBLE MANAGED BLOCK commander
+      [Production]
```
dav3r: Not sure if this matters, but shouldn't this be lower-case 'p' `[production]` to match the contents of the `block` below?

jsf9k: Yes.  I'm not sure how that happened.  Fixing now.

KyleEvers (approved):
👍

dav3r (approved):
Strong work all! 🤖

dav3r:
Note that this PR resolves issue #129

PR #163, opened 2019-01-29
Improvement/use nessus 8

dav3r:
These are the changes required by CyHy in order to use Nessus Professional 8.x.

**IMPORTANT: These changes should NOT be merged into the main develop branch until our Nessus licensing issues have been sorted out (hence the "Work in progress" label).**

This change is necessitated by the [Nessus Service Bulletin](https://docs.tenable.com/releasenotes/nessus/servicebulletin.htm) which explains that older versions of Nessus (prior to 7.1.3) will no longer be able to compile the plugin database and run scans after the plugin database size exceeds 2 GB, an event which is expected to happen soon after Jan. 31, 2019.

NOTES: 
- This code requires the Nessus 8.x Debian package to be available in an S3 bucket.  That package can be downloaded [here](https://www.tenable.com/downloads/nessus).  In the future, it would be preferable to automate the process of fetching the package directly from Tenable.
- The CyHy base scan policy XML previously used with Nessus 6.x was determined to be incompatible with Nessus 8.x, so a fresh version of the CyHy base scan policy was created with Nessus 8 and stored in S3.


felddy (approved):
So long Amazon Linux - Strong Work DaveR!

PR #168, opened 2019-02-04
Bugfix/feeds cronjob only in prod

KyleEvers:


jsf9k (approved):
Looks good to me!
:eye: :eye:

PR #169, opened 2019-02-04
Move creation of the Mongo run directory from the second Ansible to the Mongo unit file

jsf9k:
The second Ansible does not run on a reboot, so on a reboot the run directory does not get created and the MongoDB service cannot start.  `systemd` provides the `RuntimeDirectory` directive for the unit file for exactly this situation, so it makes sense to use it here.

I tested in the jsf9k Terraform workspace that this works and MongoDB automatically restarts after a reboot.


KyleEvers (approved):
Looks good 👍

PR #171, opened 2019-02-07
A bit of cleanup

jsf9k:
* Remove the reporter instance's explicit dependency on the database instance.
* Add volume tags for the reporter.


dav3r (approved):
🥑

PR #174, opened 2019-02-15
Change public zones from cyber.dhs.gov to ncats.cyber.dhs.gov.

jsf9k:
This is a change that @felddy asked for.  The analogous change has already been made in the [dhs-ncats/pca-teamserver-aws](https://github.com/dhs-ncats/pca-teamserver-aws) repository.

This change is simple, but it will likely break folks' local ssh configs.  It will also likely break the CyHy team's access to the Mongo DB in AWS.  Some coordination will be required before deployment.

### DIFF
--- a/terraform/locals.tf
+++ b/terraform/locals.tf
@@ -48,10 +48,10 @@ locals {
   bod_private_domain = "local"
 
   # zone to use for public DNS
-  cyhy_public_zone  = "cyber.dhs.gov"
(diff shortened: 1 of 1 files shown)


KyleEvers (approved):
Highway to the danger zone 🚗 💀

felddy (approved):
Thanks Shane.  Looks good.

PR #178, opened 2019-02-25
Create Management VPC

dav3r:
This PR adds a new "management" VPC that contains a bastion instance and a vulnscan (Nessus) instance.  A new variable called `enable_mgmt_vpc_access_to_all_vpcs` controls whether or not the management VPC is able to communicate with the other VPCs (CyHy and BOD).  Its value should be set to 1 when access is needed (e.g. during security vulnerability scans of the CyHy/BOD/Management VPCs) and set to 0 when access is not needed (all other times).

This feature was requested in CYHY-714.


terraform/bod_bastion_security_group_rules.tf
```diff
@@ -34,3 +34,93 @@ resource "aws_security_group_rule" "bastion_ssh_to_docker" {
...
+resource "aws_security_group_rule" "bod_bastion_ingress_all_icmp_from_mgmt_vulnscan" {
+  count = "${var.enable_mgmt_vpc_access_to_all_vpcs}"
+
+  security_group_id = "${aws_security_group.bod_bastion_sg.id}"
+  type = "ingress"
+  protocol = "icmp"
+  cidr_blocks = [
+    "${aws_instance.mgmt_nessus.private_ip}/32"
+  ]
+  from_port = -1
```
KyleEvers, line 49: Why port -1?

jsf9k (approved):
I took a look at this back in the jurassic, and it looked good then.  I looked at the commits since then and they look good to me as well.

KyleEvers (approved):
🥇 Looks good

dav3r:
One other thought that I want to put out there for comment:

Does it make more sense to have the variable (currently called `enable_mgmt_vpc_access_to_all_vpcs`) also control whether or not instances within the management VPC are actually created?  So if the variable were set to 0, the bastion and vulnscan instances in the management VPC would be destroyed, saving us some money.  The only downside I can think of is the Nessus license- it would need to be reset every time the vulnscan instance was destroyed and we can only reset each license every 10 days.

jsf9k:
I like that idea, and I assumed we were not doing that already simply because of the Nessus license issue.  At a minimum the bastion's existence could still be dependent on that variable.

terraform/bod_bastion_security_group_rules.tf
dav3r: To close the loop on this, when the protocol is ICMP, -1 for `from_port` and `to_port` means "allow all types of ICMP".  I cannot find any official documentation on this, but a web search turns up various posts that mention it, like this:  https://blog.jwr.io/terraform/icmp/ping/security/groups/2018/02/02/terraform-icmp-rules.html

dav3r:
Sounds good- for now, I'm going to go ahead and make the bastion dependent on that variable, but not the vulnscan instance.

KyleEvers:
I like the idea also. Destroying the Nessus instance is more important than the bastion if cost saving is the goal, but I don't know the frequency in which we will be using Nessus to scan.

jsf9k:
The bastion can be easily destroyed and recreated, so we may as well save a few dollars there if we can, but I agree that the real savings is with the Nessus instance.

dav3r:
The annoying bit is going to be making the provisioner module dynamically created, since we can't use the terraform `count` feature with modules.

jsf9k:
Right.  I guess you can do what @felddy did for the nmap and Nessus scanners, or just leave it as it is for now.

dav3r:
I'm going to leave it as is for now and think about it a bit more.  Once the management VPC starts getting used regularly, I'm hoping we will have a better feel for what we should do here.

ansible/roles/mgmt_ops/meta/main.yml
```diff
@@ -0,0 +1,57 @@
...
+  # issue_tracker_url: http://example.com/issue/tracker
+
+  # Some suggested licenses:
+  # - BSD (default)
+  # - MIT
+  # - GPLv2
+  # - GPLv3
+  # - Apache
+  # - CC-BY
+  license: license (GPLv2, CC-BY, etc)
```
felddy: I know this is probably auto-generated but do we want it to be set to `CC0`?

felddy (approved):
Crikey.  Looks good.   Might need some docs on how we _mix in_ the correct sauce in future terraforms so that we get the proper coordination with the management VPC.  This is a lot to ingest.   Great job!

ansible/roles/mgmt_ops/meta/main.yml
jsf9k: [That's exactly what I did in the Ansible roles that I broke out into their own repos](https://github.com/dhs-ncats/ansible-role-htop/blob/develop/meta/main.yml#L6).  We should probably do it everywhere (and get rid of the autogenerated cruft that we don't need).

I need to get back to breaking more roles out into their own repos.  Once I get my backlog of PRs from the shutdown shepherded through I'll make some time for that.

dav3r: I can change it here since it relates to this PR, but it should also be done for all of the ansible roles that we have in cyhy_amis (or as @jsf9k said, just do it when they get split out into their own repos).

dav3r: License added in [this commit](https://github.com/dhs-ncats/cyhy_amis/pull/178/commits/84c397f4e2b09e6f05edb203633d3d012fed4755)

PR #179, opened 2019-02-28
Pull Lambda zips from an S3 bucket instead of from the local filesystem

jsf9k:
I am making this enhancement because then I can upload the correct zips to the appropriate bucket.  Then other users need not worry about generating the zips themselves in another repo, or making sure that the zips end up in the correct location.  This makes life easier for everyone.

Note that Terraform cannot access buckets that are not in the provider's region.  This limitation means that we have to create region-specific buckets.

This PR is also a bugfix since it corrects a couple of minor issues that somehow made it through #153 and/or #154.


dav3r (approved):
Looks good!  ⚡️

PR #180, opened 2019-03-01
Fix some bugs discovered when applying to production

jsf9k:
* Remove some Google DNS cruft from `bod_public_acl_rules.tf`.
* Remove `depends_on` and `tags` from `bod_production_eip` since it's just a data element.


felddy (approved):
Less is more.  Nice!

PR #181, opened 2019-03-01
change url that feeds is pulled from and remove github auth

KyleEvers:


jsf9k (approved):
Strong work, @KyleEvers!

PR #183, opened 2019-03-04
Increase jobs-per-nessus-host from 16 to 32

dav3r:


jsf9k (approved):
Stunning work as always, @dav3r.

PR #185, opened 2019-03-06
Remove explicit touch command

jsf9k:
We can instead use the `file` Ansible module and set the state parameter equal to `touch`.

When we use the `command` Ansible module to run `touch` it generates a warning, which has been slowly driving me mad.  Today I finally got around to fixing this.


dav3r (approved):
How touching. 🐻

PR #186, opened 2019-03-11
Allow the user to specify docker-compose override files

jsf9k:
`docker-compose` by default reads in the `docker-compose.yml` and `docker-compose.override.yml` files.  By symlinking the proper YML file to `docker-compose.override.yml` in the second Ansible, we can install the proper configuration for the docker and reporter EC2 instances.  In particular, the docker instance will send the BOD 18-01 reports and the reporter instance will send the Cyber Hygiene reports and the Cyber Exposure scorecard.

See also cisagov/cyhy-mailer#42, where I created the override files specific to the cyhy and docker EC2 instances.

### DIFF
--- a/ansible/roles/cyhy_mailer/tasks/main.yml
+++ b/ansible/roles/cyhy_mailer/tasks/main.yml
@@ -66,6 +66,20 @@
       [default]
       region = {{ ses_aws_region }}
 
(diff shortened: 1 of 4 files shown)


dav3r (approved):
I like it - nice work! 🦆

PR #187, opened 2019-03-14
Improvement/more dynamic mgmt vpc

dav3r:
This PR gives us better control over the management VPC.  The bastion and vulnscanner are now handled dynamically, similar to how we support a dynamic number of scanners in the CyHy VPC.

Even though the management bastion is created dynamically, **do not** create more than one in the management VPC; the management bastion is only created dynamically so that it can be removed when the management VPC is not needed.

To completely remove the management VPC and all of its resources, do the following:
1. Set the `enable_mgmt_vpc` variable (previously known as `enable_mgmt_vpc_access_to_all_vpcs`) in your `.tfvars` file to 0
1. Update `configure.py` so that your terraform workspace is configured with:
 `'mgmt_bastion': 0, 'mgmt_nessus': 0`
1. Run `configure.py` to update your local configuration
1. Terraform away!

To create (enable) the management VPC and all of its resources, do the following:
1. Set the `enable_mgmt_vpc` variable (previously known as `enable_mgmt_vpc_access_to_all_vpcs`) in your `.tfvars` file to 1
1. Update `configure.py` so that your terraform workspace is configured with:
 `'mgmt_bastion': 1, 'mgmt_nessus': 1` (if desired, you can increase the number of `mgmt_nessus` instances, but the number `mgmt_bastion` instances should **never be more than 1**)
1. Run `configure.py` to update your local configuration
1. Ensure that the `mgmt_nessus_activation_codes` variable in your `.tfvars` file has a valid activation code for each `mgmt_nessus` instance
1. Terraform away!

### DIFF
--- a/terraform/bod_bastion_security_group_rules.tf
+++ b/terraform/bod_bastion_security_group_rules.tf
@@ -38,7 +38,7 @@ resource "aws_security_group_rule" "bastion_ssh_to_docker" {
 # Allow all ICMP from vulnscanner instance in Management VPC,
 # for internal scanning
 resource "aws_security_group_rule" "bod_bastion_ingress_all_icmp_from_mgmt_vulnscan" {
-  count = "${var.enable_mgmt_vpc_access_to_all_vpcs}"
+  count = "${var.enable_mgmt_vpc}"
 
   security_group_id = "${aws_security_group.bod_bastion_sg.id}"
   type = "ingress"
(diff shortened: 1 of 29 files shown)


jsf9k (approved):
Strong work sir

dav3r:
Ack- thanks for catching those missing braces @jsf9k - that's what I get for making late-breaking changes after I finish testing.

PR #190, opened 2019-03-27
Allow Amazon certs in CAA record

jsf9k:
I am giving Amazon certs a try for `rules.ncats.cyber.dhs.gov` instead of Let's Encrypt.  I suspect this will make certificate renewal easier.

I had to add two records to our Route53 configuration in order to accomplish this.

### DIFF
--- a/terraform_route_53/cyber.dhs.gov.tf
+++ b/terraform_route_53/cyber.dhs.gov.tf
@@ -36,6 +36,7 @@ resource "aws_route53_record" "root_CAA" {
   type    = "CAA"
   ttl     = 300
   records = [ "0 issue \"letsencrypt.org\"",
+              "0 issue \"amazon.com\"",
               "0 issuewild \";\"",
               "0 iodef \"mailto:NCATSSecurity@hq.dhs.gov\""
             ]
--- a/terraform_route_53/cyber.dhs.gov.tf
+++ b/terraform_route_53/cyber.dhs.gov.tf
@@ -42,6 +42,18 @@ resource "aws_route53_record" "root_CAA" {
             ]
 }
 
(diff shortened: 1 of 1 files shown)


jsf9k:
I went ahead and updated the `rules.ncats.cyber.dhs.gov` content as well.  We should really look into automating that.

terraform_route_53/cyber.dhs.gov.tf
```diff
@@ -36,11 +36,24 @@ resource "aws_route53_record" "root_CAA" {
   type    = "CAA"
   ttl     = 300
   records = [ "0 issue \"letsencrypt.org\"",
+              "0 issue \"amazon.com\"",
               "0 issuewild \";\"",
               "0 iodef \"mailto:NCATSSecurity@hq.dhs.gov\""
             ]
 }
 
+# This DNS record gives Amazon Certificate Manager permission to
```
jsf9k, line 45: @felddy, I wasn't sure if you'd prefer this CNAME record to be here or [here](https://github.com/cisagov/cyhy_amis/blob/develop/terraform_egress_pub/cloudfront.tf).

jsf9k:
In the interest of full disclosure, I feel I should mention that this code is already deployed to the `prod-a` workspace, and the rules site is up and running with an Amazon cert now.

terraform_route_53/cyber.dhs.gov.tf
felddy: Either place makes sense.  I'd leave it where it is for now.  Thanks for asking.

felddy (approved):
Looks good.  Let's not forget how to use `Let's Encrypt`... tying our horse to AWS, yada yada yada...  you know my schtick.

PR #191, opened 2019-03-30
Add code-gov-update to BOD Docker instance

jsf9k:
I want to eventually use ECS for this, but until I figure out the details for that I will run this via a cron job from the BOD Docker instance.


dav3r (approved):
Looking good @jsf9k 🍰

PR #193, opened 2019-04-13
Add client cert update

jsf9k:
Add Ansible roles for [client-cert-update](https://github.com/cisagov/client-cert-update).

See NCATS JIRA ticket CYHYDEV-757 for more details.

### DIFF
--- a/ansible/roles/chown_cyhy_dirs/tasks/main.yml
+++ b/ansible/roles/chown_cyhy_dirs/tasks/main.yml
@@ -11,5 +11,5 @@
   loop:
     - /var/cyhy
     - /var/log/cyhy
-  async: 180
+  async: 500
   poll: 10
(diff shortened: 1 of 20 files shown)


ansible/roles/chown_cyhy_dirs/tasks/main.yml
```diff
@@ -11,5 +11,5 @@
   loop:
     - /var/cyhy
     - /var/log/cyhy
-  async: 180
+  async: 500
```
dav3r, line 14: Did we have a case where this `chown` step was taking more than 3 minutes to complete?

dav3r (approved):
Approved, but please see the question I asked.

ansible/roles/chown_cyhy_dirs/tasks/main.yml
jsf9k: Yes, it was sporadically timing out when I applied to my `jsf9k` workspace.  Not often, but occasionally.

PR #194, opened 2019-04-15
Add workspace support to scripts

jsf9k:
Now you can specify the workspace on the command line, or not.  If you don't specify a workspace then the `prod-a` workspace is used.

### DIFF
--- a/terraform/deploy_new_docker_ami.sh
+++ b/terraform/deploy_new_docker_ami.sh
@@ -1,8 +1,19 @@
 #!/usr/bin/env bash
 
+# deploy_new_docker_ami.sh
+# deploy_new_docker_ami.sh workspace_name
+
 set -o nounset
 set -o errexit
 set -o pipefail
+
+workspace=prod-a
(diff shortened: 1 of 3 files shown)


felddy:
Looks good.  Couple of ideas.  
- Should we document these support scripts in the [`terraform/README.md`](https://github.com/cisagov/cyhy_amis/blob/develop/terraform/README.md)?
- Should we differentiate these scripts from the ones living in the sibling [`scripts`](https://github.com/cisagov/cyhy_amis/tree/develop/terraform/scripts) folder?
  - is a `support_scripts` folder warranted?

felddy (approved):

jsf9k:
I don't think we need to document these scripts in the `terraform/README.md` or anywhere else.  They are provided solely for our convenience.  I also think they can just live where they are.

In The Future (TM) I hope to have an automated system for applying updates when warranted by CI/CD and at a minimum weekly when there are no code changes.  At that time these scripts will be OBE.

What do you think, @dav3r and @KyleEvers?

PR #197, opened 2019-04-19
Docker ansible role separation

KyleEvers:
Pulls docker out from being a single ansible role for cyhy_amis to having a separate repository to be used across multiple projects.


jsf9k (approved):
This looks good to me.

PR #198, opened 2019-04-19
Move python and pip Ansible roles to their own repos

jsf9k:
See [cisagov/ansible-role-python](https://github.com/cisagov/ansible-role-python) and [cisagov/ansible-role-pip](https://github.com/cisagov/ansible-role-pip).


felddy (approved):
pip pip cheerio!

PR #199, opened 2019-04-20
Allow DNS resolution across the CyHy and BOD VPCs

jsf9k:
Give the various VPCs unique local domains instead of using `.local` for all three.  This allows us to:

1. Resolve `database1.cyhy` from the Docker EC2 instance in the BOD VPC, for example.
2. Stop passing the private IP of the MongoDB instance into the second Ansible as a parameter.


jsf9k:
@dav3r, do you want this kind of DNS resolution for the management VPC?  If so I can add it in a conditional way that depends on `enable_mgmt_vpc` by using [this resource type](https://www.terraform.io/docs/providers/aws/r/route53_zone_association.html).

dav3r:
@jsf9k, yes it will be useful to have DNS resolution of the CyHy and BOD hosts in the management VPC, mainly for when we set up the list of targets to be vulnerability scanned by the scanner in the management VPC.

dav3r:
Not sure if it really needs to be updated, but perhaps for consistency, we should update `vulnscan1` and `database1` to `vulnscan1.cyhy` and `database1.cyhy` in the [SSH port-forwarding section of the terraform README](https://github.com/cisagov/cyhy_amis/blob/develop/terraform/README.md#ssh-port-forwarding).

dav3r (approved):
Everything here is gravy. 😋

jsf9k:
> Not sure if it really needs to be updated, but perhaps for consistency, we should update `vulnscan1` and `database1` to `vulnscan1.cyhy` and `database1.cyhy` in the [SSH port-forwarding section of the terraform README](https://github.com/cisagov/cyhy_amis/blob/develop/terraform/README.md#ssh-port-forwarding).

That shouldn't be necessary.  For hosts in the CyHy VPC, for example, the lookup adds `.cyhy` by default if necessary.  (Previously it was doing the same thing, but adding `.local`.)

jsf9k:
I'm adding the `mgmt` resolution changes now @dav3r.  I'm just testing to make sure it works whether the management VPC is enabled or not.

jsf9k:
Let me know if you're still happy with this, @dav3r.  If you are then I'll merge.  Everything worked perfectly in my testing.

dav3r:
Yup, that looks good to me.  @jsf9k did you test the hostname resolution in your local workspace, including the management VPC?

jsf9k:
> Yup, that looks good to me. @jsf9k did you test the hostname resolution in your local workspace, including the management VPC?

Yessir, I did!

PR #201, opened 2019-04-26
Move secrets from S3 to SSM

jsf9k:
This allows us to reference the OAUTH token via the inline `aws_ssm` Ansible lookup instead of having to download the file to `/tmp` from S3, import its contents into Ansible as a variable, and delete the `/tmp` file.

This means that we can get rid of the [cisagov/ansible-role-github-oauth](https://github.com/cisagov/ansible-role-github-oauth) Ansible role and GitHub repo, for example.  In other Ansible roles it leads to a reduction in the amount of Ansible code.  It is also much less likely that a `/tmp` file could leak out by being left on a dev's local machine.

[See here](https://docs.ansible.com/ansible/latest/plugins/lookup/aws_ssm.html) for more details about the `aws_ssm` Ansible lookup.

### DIFF
--- a/ansible/roles/cyhy_commander/vars/main.yml
+++ b/ansible/roles/cyhy_commander/vars/main.yml
@@ -1,2 +1,15 @@
 ---
 # vars file for cyhy_commander
+
+# The CyHy SSH private key
+ssh_private_key: "{{ lookup('aws_ssm', '/cyhy/ssh/private_key') }}"
+
+# commander mongo username
+commander_user: "{{ lookup('aws_ssm', '/cyhy/mongo/users/commander/user') }}"
+# commander mongo password
+commander_pw: "{{ lookup('aws_ssm', '/cyhy/mongo/users/commander/password') }}"
+# commander mongo database
+commander_db: "{{ lookup('aws_ssm', '/cyhy/mongo/users/commander/database') }}"
+
+# The GitHub OAUTH token
+oauth_token: "{{ lookup('aws_ssm', '/github/oauth_token') }}"
--- a/ansible/roles/cyhy_commander/vars/main.yml
+++ b/ansible/roles/cyhy_commander/vars/main.yml
@@ -12,4 +12,4 @@ commander_pw: "{{ lookup('aws_ssm', '/cyhy/mongo/users/commander/password') }}"
 commander_db: "{{ lookup('aws_ssm', '/cyhy/mongo/users/commander/database') }}"
 
 # The GitHub OAUTH token
-oauth_token: "{{ lookup('aws_ssm', '/github/oauth_token') }}"
+github_oauth_token: "{{ lookup('aws_ssm', '/github/oauth_token') }}"
--- a/ansible/roles/cyhy_feeds/vars/main.yml
+++ b/ansible/roles/cyhy_feeds/vars/main.yml
@@ -1,2 +1,36 @@
 ---
 # vars file for cyhy_feeds
+
+# cyhy-feeds config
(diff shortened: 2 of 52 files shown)


jsf9k:
I've moved several key pieces of data from S3 to SSM already.  I wanted to get buy-in from the other @cisagov/team-ois members before proceeding down this path.  Please let me know what you think!

jsf9k:
@dav3r and @felddy, is our Nessus config actually a secret or could it be included as a file with the Nessus Ansible role?

felddy:
The Nessus config is not secret, and can live in the Ansible role.

ansible/roles/cyhy_commander/vars/main.yml
```diff
@@ -1,2 +1,15 @@
...
+
+# commander mongo username
+commander_user: "{{ lookup('aws_ssm', '/cyhy/mongo/users/commander/user') }}"
+# commander mongo password
+commander_pw: "{{ lookup('aws_ssm', '/cyhy/mongo/users/commander/password') }}"
+# commander mongo database
+commander_db: "{{ lookup('aws_ssm', '/cyhy/mongo/users/commander/database') }}"
+
+# The GitHub OAUTH token
+oauth_token: "{{ lookup('aws_ssm', '/github/oauth_token') }}"
```
dav3r: I'd vote for a slightly more descriptive name here, like `github_oauth_token`

ansible/roles/cyhy_feeds/vars/main.yml
```diff
@@ -1,2 +1,36 @@
...
+
+# cyhy-feeds public GPG key
+public_gpg_key: "{{ lookup('aws_ssm', '/cyhy/feeds/gpg/public') }}"
+# cyhy-feeds private GPG key
+private_gpg_key: "{{ lookup('aws_ssm', '/cyhy/feeds/gpg/private') }}"
+# NSD public GPG key
+nsd_public_gpg_key: "{{ lookup('aws_ssm', '/cyhy/feeds/gpg/nsd_public') }}"
+
+# cyhy-feeds GPG trust
+trust: "{{ lookup('aws_ssm', '/cyhy/feeds/gpg/trust') }}"
```
dav3r: Same here, I'd prefer to see this called `gpg_trust` to improve readability

dav3r (approved):
I'm fully on board with the plan to move to AWS SSM.  It really does simplify things for us and has no discernible downsides.

dav3r:
@jsf9k, @felddy Regarding the Nessus policy, the reason @KyleEvers initially made this a secret was because our Nessus username is displayed in the `<policyOwner>` data element.  We were concerned about this bit of data leakage, so we took a more cautious approach.  Perhaps we can store the username as a secret and substitute it into the XML file as a template?

ansible/roles/cyhy_feeds/vars/main.yml
jsf9k: Changed in 5a7097c.

ansible/roles/cyhy_commander/vars/main.yml
jsf9k: Renamed in f83ed1f.

jsf9k:
> @jsf9k, @felddy Regarding the Nessus policy, the reason @KyleEvers initially made this a secret was because our Nessus username is displayed in the `<policyOwner>` data element. We were concerned about this bit of data leakage, so we took a more cautious approach. Perhaps we can store the username as a secret and substitute it into the XML file as a template?

This change was made in 0a094e4.

jsf9k:
After @dav3r was able to get me a trial key, I was able to successfully test the CyHy Nessus instance.

PR #203, opened 2019-05-24
Move Packer Ansible roles to their own GitHub repos

jsf9k:
All the Ansible roles used when Packer creates the AMIs have now been moved to their own GitHub repos and configured to use [Molecule](https://github.com/ansible/molecule) for testing.

I verified that I could build all existing AMIs with Ansible 2.8.0.  I suspect that a recent version of Ansible 2.7 would work as well, and possibly even a recent version of Ansible 2.6.


dav3r (approved):
This all looks great to me!  ⚙️ 💎 🎉

PR #204, opened 2019-05-29
Add cron job on reporter to create and email daily CyHy notifications

dav3r:
Note that this PR is dependent on cisagov/ansible-role-cyhy-reports#3.


jsf9k (approved):
Looks good!

PR #207, opened 2019-05-31
Add runbook: 'How to Redeploy All Instances'

dav3r:


felddy (approved):
I like that it lives in the repo now.  
Might we add how to terminate all the EC2 instances using the AWS CLI instead of the console?

terraform/runbooks/redeploy_instances.md
```diff
@@ -0,0 +1,68 @@
...
+* Notify all affected parties that downtime is starting
+
+* ssh to the `database1` instance and stop the `cyhy-commander` service:
+
+  ```bash
+  sudo systemctl stop cyhy-commander.service
+  ```
+
+* Terminate all EC2 instances through the
+  [AWS Console](https://aws.amazon.com/console/)
```
felddy, line 25: What is the CLI command for this?

dav3r: It's `aws ec2 terminate-instances --instance-ids <id1> <id2> ...`  

Here's a handy snippet to get all of the instance IDs in a particular terraform workspace (assuming all of the instances are properly tagged with the workspace name):
`aws ec2 describe-instances --filters "Name=tag:Workspace,Values=<WORKSPACE_NAME>" --output text --query "Reservations[*].Instances[*].InstanceId"`

So one way to do it would be:
`aws ec2 describe-instances --filters "Name=tag:Workspace,Values=<WORKSPACE_NAME>" --output text --query "Reservations[*].Instances[*].InstanceId" | xargs aws ec2 terminate-instances --instance-ids
`

Our goal should be to not have to do this step at all.  Hopefully, after we move to terraform 0.12, we will be able to clean up our code so that `terraform apply` will be able to cleanly terminate our instances, then spin up their replacements (or in certain cases, we may be able create the new instances before destroying the old ones; eventually, we will want that model across the board).

felddy: Roger.  Terraform 0.12 will fix all our woes!  ;)

PR #208, opened 2019-06-04
Get rid of some Ansible warnings

jsf9k:
Ansible 2.8 is touchier about types and therefore outputs warnings about:
* Not casting the production_workspace variable to a bool
* Using integer values where a string value is expected

This pull request makes some minor adjustments so that these warnings go away.


dav3r (approved):
LGTM! 👍 

I will update the cyhy_reporter notifications cron job similarly in my own PR.

PR #209, opened 2019-06-04
Update script location for daily CyHy notification cron job

dav3r:
The location of the daily notification script (`create_send_notifications.py`) changes in cisagov/ansible-role-cyhy-reports#4.  This PR updates the cron job to point to the new location.

This PR should be deployed in conjunction with cisagov/ansible-role-cyhy-reports#4 and cisagov/cyhy-mailer#54.

### DIFF
--- a/ansible/roles/cyhy_reporter/tasks/main.yml
+++ b/ansible/roles/cyhy_reporter/tasks/main.yml
@@ -64,7 +64,7 @@
     minute: 0
     hour: 10
     user: cyhy
-    job: cd /var/cyhy/notifications && ./create_send_notifications.py --log-level=info cyhy 2>&1 | /usr/bin/logger -t cyhy-notifications
+    job: cd /var/cyhy/reports && ./create_send_notifications.py --log-level=info cyhy 2>&1 | /usr/bin/logger -t cyhy-notifications
   when: production_workspace
 
 #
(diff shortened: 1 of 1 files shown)


ansible/roles/cyhy_reporter/tasks/main.yml
```diff
@@ -61,10 +61,10 @@
...
   cron:
     name: "Generate and send daily CyHy notifications"
-    minute: 0
-    hour: 10
+    minute: '0'
+    hour: '10'
     user: cyhy
-    job: cd /var/cyhy/notifications && ./create_send_notifications.py --log-level=info cyhy 2>&1 | /usr/bin/logger -t cyhy-notifications
+    job: cd /var/cyhy/reports && ./create_send_notifications.py --log-level=info cyhy 2>&1 | /usr/bin/logger -t cyhy-notifications
   when: production_workspace
```
jsf9k: This should be `production_workspace|bool` to avoid an Ansible 2.8 warning.

jsf9k (changes requested):
Requesting one minor change.  Otherwise this looks good to me.

ansible/roles/cyhy_reporter/tasks/main.yml
dav3r: Good catch, thanks @jsf9k - fixed in 2678ef5

jsf9k (approved):
Huzzah!

PR #212, opened 2019-06-07
change webui service to only restart on failure and increase restart …

KyleEvers:
…time


jsf9k (approved):
Strong work, @KyleEvers

PR #213, opened 2019-06-10
Update script for Terraform 0.12

jsf9k:
Update the `deploy_new_lambdas.sh` script.  The syntax for identifying an item in an array has changed in Terraform 0.12.

Also add a script for deploying a new portscan instance.  This is occasionally useful when doing a full deploy, since sometimes one of the portscan instances fails to mount its data EBS volume and needs to be redeployed.


dav3r (approved):
LGTM!  👍 

I will make a PR soon to add the region into two `aws` commands (for this and all of the other deploy scripts).

PR #214, opened 2019-06-10
Add a workaround for a bug related to empty lists

jsf9k:
If you terminate all the existing instances for a particular template and then run `terraform apply`, the corresponding list `var.*_private_ips` is empty at that time.  Then there is an error condition when Terraform evaluates what must be done for the apply because you are trying to use `element()` to reference indices in an empty list.  The list will be populated with the actual values as the apply runs, so we just need to get past the pre-apply stage.  Therefore this ugly hack works.

If you find a better way, please use it and get rid of this affront to basic decency.


dav3r (approved):
Ugly, but it works. Hopefully this will go away when we no longer have to create dynamic modules on our own (see https://github.com/hashicorp/terraform/issues/953).

Thanks @jsf9k ! 🎖

PR #215, opened 2019-06-10
Go back to using cloudposse/terraform-null-ansible

jsf9k:
The guys at cloudposse merged [my pull request](https://github.com/cloudposse/terraform-null-ansible#23), so we no longer need to pull from my fork.


dav3r (approved):
Go party posse! 🥇

PR #216, opened 2019-06-11
Fix deploy script

jsf9k:


terraform/deploy_new_portscan_instance.sh
```diff
@@ -27,7 +27,7 @@ terraform workspace select "$workspace"
 # Strip control characters, then look for the text "id" surrounded by
 # space characters, then extract only the ID from that line.
 nmap_instance_id=$(terraform state show aws_instance.cyhy_nmap[$index] | \
-                       sed -r "s/\x1b\[[0-9;]*m?//g" | \
+                       sed $'s,\x1b\\[[0-9;]*[[:alpha:]]],,g' | \
```
dav3r: This does NOT work correctly on OSX:
`sed $'s,\x1b\\[[0-9;]*[[:alpha:]]],,g'`, however this does:
`sed $'s,\x1b\\[[0-9;]*[[:alpha:]],,g'`

Not sure where that extra `]` came from, but it wasn't there when you gave me samples to test.

dav3r (changes requested):
See my note for a slight change required to make this work with the BSD sed on OSX.

terraform/deploy_new_portscan_instance.sh
jsf9k: Good eye.  I fat fingered that.  Fixed in 803e622.

dav3r (approved):
Looking good! 🎉

PR #217, opened 2019-06-11
Deploy script improvements

dav3r:
Update all deploy scripts to:
* Require all command-line parameters; previously, scripts would default to `prod-a` workspace
* Include new 'region' parameter so we can run scripts in regions other than previous default of `us-east-1`
* Include usage output if no parameters given (for scripts that didn't already have this)


jsf9k (approved):
Looks good to me!

PR #218, opened 2019-06-11
Update splat syntax

jsf9k:
There is a new and much more intuitive version of the Terraform splat syntax in version 0.12, so I updated the code to use it.

@dav3r and I noticed some cases where we were previously setting `count` equal to `local.nessus_instance_count`, for example, and then indexing into structures such as `aws_instance.cyhy_nessus[count.index].id`.  This used to work, but now can cause problems.  This is because when `terraform apply` is run, `length(aws_instance.cyhy_nessus)` and `local.nessus_instance_count` are only _eventually_ consistent.  Therefore we sometimes get errors because we're indexing into an empty list, for example.

A better practice is to set `count` equal to the thing we're actually indexing into where possible, such as `length(aws_instance.cyhy_nessus)`.  I made that change in several places in this pull request, and we should adopt this practice going forward.

### DIFF
--- a/terraform/bod_lambdas.tf
+++ b/terraform/bod_lambdas.tf
@@ -25,7 +25,7 @@ resource "aws_iam_role" "lambda_roles" {
 # generate log output in Cloudwatch.  These will be applied to the
 # roles we are creating.
 data "aws_iam_policy_document" "lambda_cloudwatch_docs" {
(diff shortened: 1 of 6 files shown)


dav3r (approved):
This all looks good to me!  🔍

PR #220, opened 2019-06-13
Cyhy feeds/remove cyhy core dependency

mcdonnnj:
Update to create configuration files for MongoDB connections and adjusting the cron job to use updated command line syntax for changes in cyhy-feeds per https://github.com/cisagov/cyhy-feeds/pull/26

### DIFF
--- a/ansible/roles/cyhy_feeds/tasks/main.yml
+++ b/ansible/roles/cyhy_feeds/tasks/main.yml
@@ -29,8 +29,8 @@
 # Create yml files for db access
 #
 
-# commander
-- name: Create cyhy.yml to store commander db credentials
+# reporter
+- name: Create cyhy.yml to store reporter db credentials
   copy:
     dest: /var/cyhy/scripts/cyhy-feeds/cyhy.yml
     owner: cyhy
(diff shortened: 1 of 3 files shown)


KyleEvers (approved):
This looks good. Strong work

ansible/roles/cyhy_feeds/tasks/main.yml
```diff
@@ -25,6 +25,46 @@
...
+# commander
+- name: Create cyhy.yml to store commander db credentials
+  copy:
+    dest: /var/cyhy/scripts/cyhy-feeds/cyhy.yml
+    owner: cyhy
+    group: cyhy
+    mode: 0660
+    content: |
+      database:
+        uri: mongodb://{{ commander_user }}:{{ commander_pw }}@database1.cyhy:27017/{{ commander_db }}
```
dav3r: I believe (though someone should double-check this) `cyhy-feeds` only needs read-access to the CyHy database.  If that is the case, we should use the `/cyhy/mongo/users/reporter/*` credentials, rather than the `commander` credentials (fetched [here](https://github.com/cisagov/cyhy_amis/blob/develop/ansible/roles/cyhy_feeds/vars/main.yml#L17-L22)), which have write-access.

jsf9k (approved):
These changes look good to me.  You can merge once cisagov/cyhy-feeds#26 and cisagov/ansible-role-cyhy-feeds#3 are merged.

dav3r (changes requested):
Looks good @mcdonnnj, but I'd like to see one thing changed first- please take a look at the comment I left.

ansible/roles/cyhy_feeds/tasks/main.yml
jsf9k: :eagle: :eye:, @dav3r!

mcdonnnj: Pulled credentials and references are updated now, please let me know if I missed something.

dav3r (approved):
This is perfect, nice job @mcdonnnj  🥇

PR #221, opened 2019-06-13
Increase retry count in nessus_base.py

dav3r:
This PR increases the maximum number of retries for a failed request in [nessus_base.py](https://github.com/cisagov/cyhy_amis/blob/develop/ansible/roles/nessus/files/nessus_base.py#L34). 
 This will be helpful in cases when `nessus_base.py` is run shortly after the `nessusd` service has been started/restarted and it is starting up slowly (e.g. when the plugin database is being built).

Currently, the ansible provisioner often fails for this reason when a new Nessus instance is initially deployed; the goal of this PR is to remedy that sitation.


jsf9k (approved):
Let's keep an eye on this and see if it resolves the issue.

PR #222, opened 2019-06-17
Add new script for redeploying the database instance

dav3r:
This will be handy whenever a new database AMI is created.  This script is very similar to the other `deploy_new_*_ami.sh` scripts that already exist.


jsf9k (approved):
:+1: :zap:

felddy (approved):
Looks good. 
💯

PR #223, opened 2019-06-17
Fix typo in URI for scan_reader.yml configuration.

mcdonnnj:
PR to fix issues with the cyhy-feeds deploy setup after it was deployed to production.


jsf9k (approved):
:us: :eagle:

dav3r (approved):
As tasty as `mmongodb` sounds, I think your fix is the way to go. 🍪

PR #224, opened 2019-06-20
Remove explicit installation of Docker

jsf9k:
Since Docker is installed as an Ansible role dependency of the Docker composition roles that use it, there is no longer a need for a Docker group.  This also necessitates some changes to the Packer JSON files.

I'd like to remove the explicit installation of Docker in `packer/dashboard.json` too, but that doesn't appear possible at this time.  Maybe when I start to converge the first and second Ansibles it will become possible.

### DIFF
--- a/packer/ansible/playbook.yml
+++ b/packer/ansible/playbook.yml
@@ -48,22 +48,27 @@
     - nessus
     - more_ephemeral_ports
 
-- hosts: docker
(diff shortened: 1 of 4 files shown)


dav3r (approved):
Getting cleaner and tidier as we go. 🎉

felddy (approved):

PR #225, opened 2019-06-22
Encrypt root volumes and delete them on termination

jsf9k:
This applies to the root disk devices of EC2 instances started from these AMIs, as well as the disk devices used when creating the AMI via `packer`.

This will put the kibosh on the proliferation of dead snapshots in AWS.

### DIFF
--- a/packer/bastion.json
+++ b/packer/bastion.json
@@ -22,6 +22,22 @@
             "us-west-2"
         ],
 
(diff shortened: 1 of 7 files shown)


dav3r (approved):
Good call! 👍

mcdonnnj (approved):
Nice additions. 👍

PR #226, opened 2019-06-24
Specify the default KMS key for each region

jsf9k:
This fixes a bug where copied AMIs were not being encrypted.  The problem is described in hashicorp/packer#6778.

Thanks to @dav3r for pointing out the bug as well as the GitHub issue with the solution.

### DIFF
--- a/packer/bastion.json
+++ b/packer/bastion.json
@@ -21,6 +21,11 @@
             "us-west-1",
             "us-west-2"
         ],
+        "region_kms_key_ids": {
+            "us-east-1": "alias/aws/ebs",
+            "us-west-1": "alias/aws/ebs",
+            "us-west-2": "alias/aws/ebs"
+        },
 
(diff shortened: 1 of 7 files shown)


dav3r (approved):
LGTM! 👍 🔑

mcdonnnj (approved):
Teamwork makes the dream work. 👍

dav3r:
@jsf9k It turns out that the issue that prompted this PR (hashicorp/packer#6778) was fixed in packer 1.4 (see hashicorp/packer#6787).  I was running packer 1.3.3 when I observed the errors that prompted this PR.  So I upgraded to packer 1.4.1 and was able to verify that we no longer need that `region_kms_key_ids` block to successfully copy AMIs to other regions.

So unless you have any concerns, I will go ahead and create a PR to revert this PR.

PR #227, opened 2019-06-25
Revert "Specify the default KMS key for each region"

dav3r:
Reverts cisagov/cyhy_amis#226

The issue that prompted cisagov/cyhy_amis#226 (hashicorp/packer#6778) was fixed in packer 1.4 (see hashicorp/packer#6787). I was running packer 1.3.3 when I observed the errors that prompted cisagov/cyhy_amis#226.

I upgraded to packer 1.4.1 and was able to verify that we no longer need that `region_kms_key_ids` block to successfully copy AMIs to other regions.

### DIFF
--- a/packer/bastion.json
+++ b/packer/bastion.json
@@ -21,11 +21,6 @@
             "us-west-1",
             "us-west-2"
         ],
-        "region_kms_key_ids": {
-            "us-east-1": "alias/aws/ebs",
-            "us-west-1": "alias/aws/ebs",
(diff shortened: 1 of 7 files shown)


jsf9k (approved):
I re-tested and the copy worked.  Maybe I updated to packer 1.4.1 since I tested yesterday, even though I don't remember doing that. :man_shrugging:

mcdonnnj (approved):
Reversion makes sense to me.

PR #230, opened 2019-06-27
Add missing tf targets to database AMI deploy script

dav3r:
This PR adds five terraform resource targets that are needed when deploying a new database AMI/instance.


felddy (approved):
🎯

mcdonnnj (approved):
We probably need those, good catch. 👍

jsf9k:
@cisagov/team-ois 

The reason they were omitted is because these scripts were originally constructed when, for example, the database AMI needed to be deployed.  At that point we ran `terraform plan`, sifted through the changes to identify only those that were dependent on the database AMI deployment, and added them to the script.  Terraform items such as these IAM policies that rarely change would not have been picked up by this process.

I mention this as a warning; there are probably more such "gotchas" lurking around these deployment scripts we created.

PR #233, opened 2019-08-21
Remove deprecated aws_flow_log argument 

dav3r:
Get rid of "log_group_name" and use "log_destination" in its place.  Why?  Because terraform told me to:
```
Warning: "log_group_name": [DEPRECATED] use 'log_destination' argument instead
```


jsf9k (approved):
Good catch!

mcdonnnj (approved):
Terraform commands it 🤖

PR #238, opened 2019-10-02
Management VPC private DNS lookups from CyHy and BOD VPCs

dav3r:
This PR associates the management private DNS zone with the CyHy and BOD VPCs to enable lookups of private DNS entries that reside in those VPCs.

This was needed so that vulnerability scans that originate from the management VPC can include private DNS names from the CyHy and BOD VPCs.


mcdonnnj (approved):
LGTM 👍

jsf9k (approved):
Strong work

Issue #236, opened 2019-09-30
Nessus Ansible Role Will Always Stop Nessus Service


mcdonnnj:
# 🐛 Bug Report

Whenever a provisioner runs the `nessus` role the given playbook will always stop the Nessus service on the AWS instance. This would result in unnecessary downtime and delays for scans, as all in progress scans are aborted and any results are discarded if the service is stopped.

["If the Nessus instance restarts or if the service is stopped, all scan jobs will abort and the results will be discarded."](https://community.tenable.com/s/article/What-happens-if-Nessus-or-the-target-reboots-in-the-middle-of-an-active-scan)

## To Reproduce

Perform a Terraform apply that results in a provisioner running the `nessus` Ansible role.

## Expected behavior

The Nessus service should only be stopped if there is no key registered, or if the key registered does not match the key to be used. That way the service is only stopped on initial instance setup or if the key is being changed.

mcdonnnj:
This should be resolved by #239

dav3r:
Excellent work @mcdonnnj - thanks! 🎡

PR #239, opened 2019-10-08
Fix Nessus being Stopped Whenever the Terraform Nessus Role is Run

mcdonnnj:
In the case where the `nessus` role run by Terraform is re-run after initial deployment the existing role with needlessly shut down the Nessus service which will interrupt scans in progress per #236 . This PR adjusts the `nessus` role so it will only stop the service when there is no Nessus activation code registered or if the code Terraform has differs from the one already registered.

### DIFF
--- a/ansible/roles/nessus/tasks/main.yml
+++ b/ansible/roles/nessus/tasks/main.yml
@@ -1,11 +1,6 @@
 ---
 # tasks file for nessus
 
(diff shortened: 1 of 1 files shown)


dav3r (approved):
Strong work @mcdonnnj 💪 💼

jsf9k (approved):
Nice job, @mcdonnnj!

PR #241, opened 2019-10-15
Correct Nessus key check regex.

mcdonnnj:
This PR fixes an overlooked issue when checking for an existing Nessus key in the Terraform `nessus` Ansible role. It currently only matches the format for Essentials keys (used during testing), but Professional keys do not conform. It is a difference of one group of characters so the change is minor (but important!).


dav3r (approved):
👍 Thanks @mcdonnnj!
I verified that this fix works with our license codes in Production.

felddy (approved):
🧀

PR #243, opened 2019-10-30
Fix for changed functionality in Terraform 0.12.12

mcdonnnj:
These changes went live in Terraform 0.12.11, but due to a bug 0.12.12 was released the next day. This is likely a result of this fix:

> config: Always evaluate whole resources rather than instances in expressions, so that invalid instance indexes can return a useful error rather than unknown (#22846)

Which results in Terraform 0.12.12 throwing an error on the 0-index references in use. Changing to splats (*) resolved in this issue and should work in line with the desired functionality.


dav3r (approved):
Thanks for fixing this @mcdonnnj! 🌎

KyleEvers (approved):
Nice workaround 👍

jsf9k (approved):
:smiley_cat:

PR #244, opened 2019-10-31
Add SSM policy for third-party CyHy reports

dav3r:
This PR modifies the instance role for the CyHy reporter so that it can read a newly-added SSM parameter (`/cyhy/report/third_party_reports`).  That SSM parameter is used to specify the list of third-party organizations that will receive CyHy reports containing data from CyHy organizations that have requested that their results be shared with the third-party.


mcdonnnj (approved):
LGTM ✅

dav3r:
Reminder- this PR may not be needed if we decide to go with the `CYHY_THIRD_PARTY` plan...

dav3r:
This PR is no longer needed.  A simpler method (one that does not require SSM) will be used to build the list of third-party report IDs.

PR #246, opened 2019-12-05
Add deploy_new_adi_lambda script

dav3r:
# 

## 🗣 Description

This PR adds a new utility script `deploy_new_adi_lambda.sh`, which should be used to redeploy the ADI lambda-related resources after a new lambda zip file has been uploaded to assessment_data_import S3 bucket.

This PR also contains a commit (https://github.com/cisagov/cyhy_amis/commit/a404dfe25c0c9b652be705f952a838bd45c72de8) that makes a small syntax change in several files to avoid triggering the following terraform 0.12.x warning:
```
Warning: Quoted type constraints are deprecated
   3:   type    = "list"

Terraform 0.11 and earlier required type constraints to be given in quotes,
but that form is now deprecated and will be removed in a future version of
Terraform. To silence this warning, remove the quotes around "list" and write
list(string) instead to explicitly indicate that the list elements are
strings.
```

## 💭 Motivation and Context

I updated the ADI lambda code in https://github.com/cisagov/assessment-data-import-lambda/pull/3 and @jsf9k made the excellent suggestion that a helper script (similar to [deploy_new_lambdas.sh](https://github.com/cisagov/cyhy_amis/blob/develop/terraform/deploy_new_lambdas.sh)) would be useful, so here it is.

## 🧪 Testing

The new script was tested and no issues were observed.

## 📷 Screenshots (if appropriate)
N/A

## 🚥 Types of Changes

## ✅ Checklist

- [ ] I have added tests to cover my changes -> Note: No new tests are needed for this PR

### DIFF
--- a/terraform/deploy_new_adi_lambda.sh
+++ b/terraform/deploy_new_adi_lambda.sh
@@ -1,17 +1,16 @@
 #!/usr/bin/env bash
 
-# deploy_new_adi_lambda.sh region workspace_name
+# deploy_new_adi_lambda.sh workspace_name
 
 set -o nounset
 set -o errexit
 set -o pipefail
 
-if [ $# -eq 2 ]
+if [ $# -eq 1 ]
 then
-    region=$1
-    workspace=$2
+    workspace=$1
 else
-    echo "Usage:  deploy_new_adi_lambda.sh region workspace_name"
+    echo "Usage:  deploy_new_adi_lambda.sh workspace_name"
     exit 1
 fi
 
--- a/terraform/deploy_new_adi_lambda.sh
+++ b/terraform/deploy_new_adi_lambda.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+# deploy_new_adi_lambda.sh region workspace_name
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+if [ $# -eq 2 ]
+then
+    region=$1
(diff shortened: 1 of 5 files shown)


terraform/deploy_new_adi_lambda.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# deploy_new_adi_lambda.sh region workspace_name
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+if [ $# -eq 2 ]
+then
+    region=$1
```
jsf9k: I don't think `region` is ever used.  It can be removed.

jsf9k (changes requested):
See comment regarding an used variable in your bash script.

image

terraform/deploy_new_adi_lambda.sh
dav3r: Ha, good catch, thanks! See https://github.com/cisagov/cyhy_amis/pull/246/commits/5a213187f09a1eff98626c295d135fb4ece6b201 for the change.  I was overzealous in my copy/pasting of a similar deploy script.

jsf9k: image

jsf9k (approved):

mcdonnnj (approved):
I'm digging the addition 👍
image

PR #247, opened 2019-12-06
Fix cyhy-feeds cron job

mcdonnnj:
# 

## 🗣 Description
This fixes the `cron` job for `cyhy-feeds` to start at 00:00 UTC instead of 08:15 UTC to align with the time window used when the `cyhy-data-extract` script is run.

This PR aligns with updates to the `cyhy-data-extract` script in https://github.com/cisagov/cyhy-feeds/pull/37

## 💭 Motivation and Context
Concerns were raised that the `cyhy-data-extract` script was either filtering or missing data in it's export. Upon investigation I realized that the start time for the script was over eight hours after the end of the time window being used. Since CyHy runs continuously it resulted in records being updated with new last time changed values that were past the end fo the query window. If a record had this happen repeatedly either due to happenstance or scan windows there was the possibility that it would never be included in a data extraction. This change has the script start at midnight to align with the end fo the query time window.

## 🧪 Testing
Manually modified production environment to use these new parameters for the `cron` job. Confirmed output is as expected.

## 🚥 Types of Changes

## ✅ Checklist

### DIFF
--- a/ansible/roles/cyhy_feeds/tasks/main.yml
+++ b/ansible/roles/cyhy_feeds/tasks/main.yml
@@ -97,7 +97,7 @@
   cron:
     name: Nightly cyhy extract
     hour: '0'
-    minute: '01'
+    minute: '0'
     user: cyhy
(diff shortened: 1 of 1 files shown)


jsf9k (approved):
Three huzzahs for these changes!

dav3r (approved):
Thanks for investigating this and coming up with an excellent solution @mcdonnnj!  🎸 

My only thought is that we should update the comments to mention the timing issue that led to this change, so that we don't forget (and future developers understand) why we run this at midnight.  

Also, I just noticed that the [existing comment](https://github.com/cisagov/cyhy_amis/blob/2e52defb2d7239bebddbe4bd4a1bdbd5d50621fc/ansible/roles/cyhy_feeds/tasks/main.yml#L93) needs to be updated from "0815 (UTC)" to "0000 (UTC)".

jsf9k:
> My only thought is that we should update the comments to mention the timing issue that led to this change, so that we don't forget (and future developers understand) why we run this at midnight.

Excellent point @dav3r!  I think a description of the timing issue is also in order, since it took us a while to figure that out.

dav3r (approved):
Stellar comment sir 👍

jsf9k (approved):
:1st_place_medal:

PR #261, opened 2020-04-27
Increase next-scan-limit for CyHy commander

dav3r:
# 

## 🗣 Description
This PR increases the default `next-scan-limit` in `commander.conf` from 3,072 to 8,192.

It also adds a useful GitHub `CODEOWNERS` file to this repo.

## 💭 Motivation and Context
Last week, I observed in the CyHy commander's log that the commander was unable to "catch up" on updating previously "down" hosts (a.k.a. "dark space") due to be scanned.  For the entire week, that portion of the log looked like this:
```
Updating previous "down" hosts that are now due to be scanned.
Updated 3072 "down" hosts.
```

That tells us that the commander was unable to get up-to-date in terms of the number of dark space IPs that were due to be re-scanned.  By increasing the `next-scan-limit` to 8,192, we give the commander a better chance of catching up in this regard.

The `next-scan-limit` of 3,072 predates our current AWS setup.  The database instance that we have in AWS is more than capable of handling this increase to 8,192.

## 🧪 Testing
I made this change in to the Production commander and ensured that there were no negative effects. 
Updating 8,192 host documents consistently took around 14 seconds in Production, so there is minimal risk of this configuration change causing a significant increase in the duration of each commander cycle.  For comparison, it was taking about 10 seconds on average to update 3,072 host documents.

## 🚥 Types of Changes

## ✅ Checklist


felddy (approved):
TURBO!  ⏩

jsf9k (approved):
:1234:

PR #268, opened 2020-05-21
Add Dashboard Deploy Script and Add Rule to Database Deploy Script

mcdonnnj:
# 

## 🗣 Description
This PR adds a script to facilitate deploying a new CyHy Dashboard AMI. There is also the addition of a rule that is required by the dashboard to the database deploy script.

The database deploy script also had its spacing adjusted to mirror the other deploy scripts in this repository.

## 💭 Motivation and Context
I ran into an issue with deployment today that involves the dashboard and it is one of the images that lacks a deploy script. Having a targeted deploy script is convenient when working with these instances.

The rule added to the database deploy script covers allowing the dashboard and reporter instances ingress to the database instance for a MongoDB connection.

I am uncertain if security group to security group rules need to be targeted when doing a redeployment. However, other deploy scripts have them, so I have included all relevant ones in the dashboard deploy script.

## 🧪 Testing
I used the dashboard script to do a targeted deploy after a forced instance replacement. I was able to successfully use dashboard services.

## 🚥 Types of Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (causes existing functionality to change)

## ✅ Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #269, opened 2020-06-17
Update Portscan Deployment Script to Support Sequential Instance Range

mcdonnnj:
# 

## 🗣 Description
This PR updates the `deploy_new_portscan_instance.sh` script to support a starting and ending instance range to deploy. It also fixes a typo noticed by @dav3r in #268 after it was merged.

## 💭 Motivation and Context
@dav3r mentioned that it would be convenient to be able to (re)deploy a range of portscan instances conveniently. This updates the deploy script to allow for that.

## 🧪 Testing
Just verification that the function and loop were providing expected values.

## 🚥 Types of Changes

## ✅ Checklist

### DIFF
--- a/terraform/deploy_new_dashboard_ami.sh
+++ b/terraform/deploy_new_dashboard_ami.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# deploy_new_database_ami.sh region workspace_name
+# deploy_new_dashboard_ami.sh region workspace_name
 
 set -o nounset
 set -o errexit
--- a/terraform/deploy_new_portscan_instance.sh
+++ b/terraform/deploy_new_portscan_instance.sh
@@ -1,40 +1,62 @@
 #!/usr/bin/env bash
 
 # deploy_new_portscan_instance.sh region workspace_name instance_index
+# deploy_new_portscan_instance.sh region workspace_name first_instance last_instance
 
 set -o nounset
 set -o errexit
 set -o pipefail
 
+function usage {
+    echo "Usage:"
+    echo "  ${0##*/} region workspace_name instance_index"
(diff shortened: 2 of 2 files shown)


jsf9k (approved):
This looks like a useful improvement.  Thanks for doing this.

dav3r (approved):
This is a nice improvement @mcdonnnj - thank you for taking the time to do it! 🥇 

My only quibble is that if we pass in a large range of instances, there will be a lot of `terraform apply` prompts that must be confirmed, with terraform processing time in between each one.  Did/can you look into writing this script in such a way that there would only be a single `terraform apply` step, regardless of the number of instances being redeployed?

jsf9k:
> My only quibble is that if we pass in a large range of instances, there
> will be a lot of terraform apply prompts that must be confirmed, with
> terraform processing time in between each one. Did/can you look into
> writing this script in such a way that there would only be a single terraform
> apply step, regardless of the number of instances being redeployed?

This would be a nice improvement!

mcdonnnj:
> 
> 
> This is a nice improvement @mcdonnnj - thank you for taking the time to do it! 🥇
> 
> My only quibble is that if we pass in a large range of instances, there will be a lot of `terraform apply` prompts that must be confirmed, with terraform processing time in between each one. Did/can you look into writing this script in such a way that there would only be a single `terraform apply` step, regardless of the number of instances being redeployed?

That's a good point. I have modified the script to make a single `terraform apply` call instead. Unless the instance count is quite large it will hopefully stay in the safe range for command line argument constraints.

dav3r (approved):
Very nice!!  Now we're cooking! 🧑‍🍳

jsf9k (approved):
Nice work!

PR #275, opened 2020-08-20
Explicitly set owner of authorized_keys file for SSH ops user

dav3r:
# 

## 🗣 Description
This PR explicitly sets the owner of `~cyhy_ops/.ssh/authorized_keys` to `cyhy_ops:cyhy_ops` when the file is created.

It also pins Terraform to 0.12.x and the AWS provider to 2.x until we can fully test the newer versions.

## 💭 Motivation and Context

A recent change to Ansible made `~cyhy_ops/.ssh/authorized_keys` be owned by `root:root` and unreadable by the `cyhy_ops` user when it was created.  This was preventing successful SSH connections from `cyhy_ops` users with authorized keys.

## 🧪 Testing
I verified the fix by changing the owner of `~cyhy_ops/.ssh/authorized_keys` to `root:root`, then doing a targeted apply of `module.cyhy_bastion_ansible_provisioner.null_resource.provisioner` and verified that `~cyhy_ops/.ssh/authorized_keys` ownership was correctly set to `cyhy_ops:cyhy_ops`. 

## 🚥 Types of Changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (causes existing functionality to change)

## ✅ Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #276, opened 2020-08-20
Explicitly set owner of authorized_keys file for SSH mgmt ops user

mcdonnnj:
# 

## 🗣 Description

This PR explicitly sets the owner of `~mgmt_ops/.ssh/authorized_keys` to `mgmt_ops:mgmt_ops` and the mode to `600` when the file is created. These changes mirror the ones done for the `cyhy_ops` role made in https://github.com/cisagov/cyhy_amis/pull/275

## 💭 Motivation and Context

After the aforementioned PR was merged, I realized that we'd run into a situation before where we made changes for the regular CyHy configuration, but hadn't made similar changes that would be needed for the management configuration. I checked and saw what would possibly cause the same problem for the management configuration and this fixes that possibility.

## 🧪 Testing

Assumption that the changes will work since they are based on an identical task in a previously tested PR.

## 🚥 Types of Changes

## ✅ Checklist

### DIFF
--- a/ansible/roles/mgmt_ops/tasks/main.yml
+++ b/ansible/roles/mgmt_ops/tasks/main.yml
@@ -25,4 +25,7 @@
     path: /home/mgmt_ops/.ssh/authorized_keys
     line: "{{ lookup('aws_ssm', '/ssh/public_keys/{{ item }}') }}"
     create: yes
+    owner: mgmt_ops
+    group: mgmt_ops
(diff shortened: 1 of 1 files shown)


jsf9k (approved):
To the *MAX*!!!

dav3r (approved):
Haha, I always forget about the lowly mgmt_ops user - thanks for remembering! 💎

PR #280, opened 2020-10-06
Tweak scraper configuration

### LINKED ISSUE #279: For llnl-scraper config, assign organization based on orgs
# 🚀 Feature Proposal

The [llnl-scraper config](https://github.com/cisagov/cyhy_amis/blob/develop/ansible/roles/code_gov_update/tasks/main.yml#L21-L46) should tailor the `organization` field to the long name of the owner of the `orgs` instead of defaulting to "Department of Homeland Security", which is already represented in `"agency": "DHS"`.

For the three orgs represented in the code:
* `cisagov` would be "Cybersecurity and Infrastructure Security Agency"
* `US-CBP` would be "U.S. Customs and Border Protection"
* `usdhs` would be "Department of Homeland Security" – and could possibly be removed at this point, since it no longer appears active (all repos are archived).

This could perhaps be achieved by variablizing the `organization` field, or maybe by running the scraper multiple times and concatenating the results.

## Motivation

It'd be useful for CISA to be better represented in code.gov. For example, a [search for COOL-related stuff at code.gov](https://code.gov/search?&page=1&query=cool&size=10&sort=best_match) shows DHS, not CISA, and that's not (as) cool.

(cc: @mcdonnnj)

jsf9k:
## 🗣 Description

This pull request tweaks the [LLNL/scraper](https://github.com/LLNL/scraper) configuration to fix an issue brought up by @h-m-f-t in #279.  Specifically, it removes the `organization` field from the configuration, so that the scraper will instead fill that field for each repo with the corresponding GitHub organization's name.  This will allow us to differentiate between CISA and CBP repositories in [the code.gov JSON](https://www.dhs.gov/code.json).

Fixes #279.

## 💭 Motivation and Context

We need to differentiate between CBP and CISA repositories in [the code.gov JSON](https://www.dhs.gov/code.json) for proper attribution.  Previously we were attributing everything to DHS.

## 🧪 Testing

All pre-commit hooks pass.

## 🚥 Types of Changes

## ✅ Checklist

### DIFF
--- a/ansible/roles/code_gov_update/tasks/main.yml
+++ b/ansible/roles/code_gov_update/tasks/main.yml
@@ -20,7 +20,6 @@
     content: |
       {
           "agency": "DHS",
-          "organization": "Department of Homeland Security",
           "contact_email": "do_not_email_just_use_github@dhs.gov",
 
           "permissions": {


ansible/roles/code_gov_update/tasks/main.yml
```diff
@@ -35,8 +34,7 @@
 
              "orgs": [
                 "cisagov",
-                "US-CBP",
-                "usdhs"
```
mcdonnnj: I understand why removing this is easier on our end, but I still believe we should include these repositories. I'm not sure how the JSON is used on the other end, but if it's a replace instead of a merge those archived repositories will no longer be listed on code.gov. GSA still has archived repositories there, and we're still including `cisagov` archived repositories. I don't think we should be eliminating other archived repositories.

dav3r (approved):
Looks good to me! 🦖

mcdonnnj (approved):
Makes sense. I had one comment that was an opinion more than a change.

h-m-f-t (approved):
🇺🇸

ansible/roles/code_gov_update/tasks/main.yml
jsf9k: @h-m-f-t suggested removing `usdhs`.  I'm happy to let him cast the deciding vote.  He may have more knowledge about the provenance of `usdhs` than we do.  What say you @h-m-f-t?

It's definitely a replace when the JSON is published to the web.

h-m-f-t: I think @mcdonnnj's comment is fair and it doesn't hurt to leave it in, so let's do that. 

(Yo @ab / @brodygov, we'd welcome if you have a different take on @usdhs.)

jsf9k: I did a force push to remove the commit where I removed `usdhs`.

PR #286, opened 2020-10-19
Update bastion Image to build on Debian Buster

mcdonnnj:
# 

## 🗣 Description

This PR updates the `bastion` instance AMI to build from Debian Buster instead of Debian Stretch.

## 💭 Motivation and Context

Currently our CyHy environment images are built on Debian Stretch because of a legacy requirement for Python 2. Any instances that can be migrated to Debian Buster should be for a variety of reasons. Since the `bastion` does not have any Python 2 requirements, migrating it to Buster is a straightforward update.

## 🧪 Testing

I built and deployed a `bastion` AMI in a testing environment and ensured that expected functionality worked.

## 🚥 Types of Changes

## ✅ Checklist

### DIFF
--- a/packer/ansible/playbook.yml
+++ b/packer/ansible/playbook.yml
@@ -12,6 +12,14 @@
     - banner
     - persist_journald
 
+# Any instances that are built on Debian Buster should have Python 2 removed
+- hosts: bastion
+  name: Remove Python 2
+  become: yes
+  become_method: sudo
+  roles:
+    - remove_python2
+
 # The bastion should have as little installed as possible, since it's
 # exposed to the cruel world
 - hosts: all:!bastion
--- a/packer/ansible/playbook.yml
+++ b/packer/ansible/playbook.yml
@@ -6,19 +6,9 @@
   become: yes
   become_method: sudo
   roles:
-    - upgrade
-    - python
     - nvme
     - banner
     - persist_journald
-
-# Any instances that are built on Debian Buster should have Python 2 removed
-- hosts: bastion
-  name: Remove Python 2
-  become: yes
-  become_method: sudo
-  roles:
-    - remove_python2
 
 # The bastion should have as little installed as possible, since it's
 # exposed to the cruel world
--- a/packer/bastion.json
+++ b/packer/bastion.json
@@ -44,7 +44,7 @@
         "Base_AMI_Name": "{{ .SourceAMIName }}",
         "OS_Version": "Debian Buster",
         "Release": "Latest",
-        "Team": "NCATS OIS - Development"
+        "Team": "VM Fusion - Development"
       },
       "type": "amazon-ebs"
     }
--- a/packer/bastion.json
+++ b/packer/bastion.json
@@ -1,59 +1,61 @@
 {
-    "builders": [{
-        "type": "amazon-ebs",
-        "region": "us-east-2",
-        "source_ami_filter": {
-            "filters": {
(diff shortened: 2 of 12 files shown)


jsf9k (approved):
Glad to see things start moving away from stretch.

packer/ansible/playbook.yml
```diff
@@ -12,6 +12,14 @@
     - banner
     - persist_journald
 
+# Any instances that are built on Debian Buster should have Python 2 removed
```
jsf9k: It's probably worth considering moving the upgrade and python steps to separate playbooks as is done [in cisagov/skeleton-docker](https://github.com/cisagov/skeleton-packer/blob/develop/src/packer.json#L65-L79).  The advantage is that Ansible "resets" and re-determines what version of python to use at the beginning of each provisioning step.  This way all the foo you really care about is installed under Ansible via python 3.

It probably doesn't matter much for the bastion, but it will matter a great deal for other AMIs that use the `pip` Ansible module.

packer/reporter.json
```diff
@@ -1,59 +1,75 @@
...
+  ],
+  "provisioners": [
+    {
+      "groups": [
+        "reporter"
+      ],
+      "playbook_file": "packer/ansible/upgrade.yml",
+      "type": "ansible"
+    },
+    {
```
jsf9k, line 60: The python playbook removes python2 from the AMI.  Are the reporter, nmap, nessus, mongo, docker, and dashboard AMIs all ready for such a big step?

mcdonnnj: The `python.yml` playbook has `remove_python2` separated into a standalone task (I didn't just blindly copy it). That task only applies to the `bastion` host at the moment.

mcdonnnj: Instance types will be added as they become Python 3 only, and eventually it will just be one task when they are all on Python 3 only.

packer/ansible/playbook.yml
jsf9k: Ah, I missed that.  Thanks for pointing it out.

packer/reporter.json
jsf9k: Ah, I missed that.  Thanks for pointing it out.

packer/bastion.json
```diff
@@ -1,59 +1,75 @@
...
+          "136693071363"
+        ]
+      },
+      "ssh_username": "admin",
+      "tags": {
+        "Application": "Cyber Hygiene",
+        "Base_AMI_Name": "{{ .SourceAMIName }}",
+        "OS_Version": "Debian Buster",
+        "Release": "Latest",
+        "Team": "NCATS OIS - Development"
```
dav3r: As long as we are updating our Packer JSON files, we might as well update our team tag to the slightly more modern `VM Fusion - Development` that we use elsewhere (e.g. [kali-packer](https://github.com/cisagov/kali-packer/blob/develop/src/packer.json#L54)).

This change should be made to all of our Packer files in this repo.

dav3r (approved):
This looks great to me!  Please consider my one minor suggestion.
Strong work @mcdonnnj - let the dominoes start falling to Debian Buster! 🌊

packer/bastion.json
mcdonnnj: Good catch! I did a find/replace on the repo in 3f8e4c6.

felddy (approved):
TRAVIS!

PR #287, opened 2020-10-21
Fix pymongo Version and File Permission

mcdonnnj:
# 

## 🗣 Description

This PR adds two small fixes I've needed to make for my testing code. They are
1. Change the temporary pymongo version installed from 3.5 to 3.6
1. Add explicit permissions for a file downloaded from GitHub

## 💭 Motivation and Context

When attempting to deploy a new database instance when using Ansible 2.10 I ran into the following issue:
```console
failed: [10.10.10.122] (item=assessment-read) => {"ansible_loop_var": "item", "changed": false, "item": "assessment-read", "msg": "pymongo driver version and MongoDB version are incompatible: you must use pymongo 3.6+ with MongoDB >= 3.6"}
```
and similar for each attempt to use the `mongodb_user` module. I tested that bumping the installed version to 3.6 was a safe remedy.

I noticed when doing an initial deploy that there were issues with the places collection being updated because the `ADDL_CYHY_PLACES.txt` file did not have appropriate permissions.

## 🧪 Testing

I was able to successfully deploy and use my testing environment with these changes in place.

## 🚥 Types of Changes

## ✅ Checklist

### DIFF
--- a/ansible/roles/mongo/tasks/main.yml
+++ b/ansible/roles/mongo/tasks/main.yml
@@ -6,17 +6,9 @@
   register: checkifmongousers
   changed_when: false
 
-# HACK ALERT - Part 1: Install pymongo 3.5 in order to use the Ansible
-# mongodb_user module
-#
-# Note that pymongo > 3.5 will fail right now, so we have to enforce
-# an old version:
-# * https://jira.mongodb.org/browse/SERVER-34820
-# * https://github.com/ansible/ansible/issues/38998
-#
-# The forced installation of 3.5 should be removed once this issue
-# gets resolved.
-- name: Install pymongo 3.5.  This should be changed to install the latest once that is possible.
+# HACK ALERT - Part 1:
+# Install pymongo 3.6 in order to use the Ansible mongodb_user module.
+- name: Install pymongo 3.6
   pip:
     name: pymongo
     version: '3.6'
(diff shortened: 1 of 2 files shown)


ansible/roles/mongo/tasks/main.yml
```diff
@@ -19,7 +19,7 @@
 - name: Install pymongo 3.5.  This should be changed to install the latest once that is possible.
   pip:
     name: pymongo
-    version: '3.5'
+    version: '3.6'
```
dav3r, line 14: If we are changing this, we can remove the `HACK ALERT` comment above and also update the name of this task to simply: `Install pymongo 3.6`

dav3r (changes requested):
This is good, but I have requested some small changes.

ansible/roles/mongo/tasks/main.yml
jsf9k: Even better - Can we just install the latest pymongo via `pip` and change the name to `Install pymongo`?

jsf9k (approved):
Glad to see some of this old detritus fall away.

ansible/roles/mongo/tasks/main.yml
dav3r: @mcdonnnj - Did you test installing the latest version of pymongo to see if it works?  That would be ideal, as @jsf9k suggested.

mcdonnnj: @dav3r Thanks for pointing that out, fixed in a2b2af0. @jsf9k Unfortunately since there is already a version of `pymongo` installed when this step runs, it appears that if we do not specify an explicit version it will not change. I tested to verify, and without a version it fails in the same way that prompted this change.

dav3r: Got it- thanks for verifying that and explaining it!

dav3r (approved):
Aww yeah! 🎉

PR #289, opened 2020-11-04
Fix portscan Instance Deploy Script

mcdonnnj:
#  #

## 🗣 Description ##

This PR fixes a bug with the `deploy_new_portscan_instance.sh` script and improves the functionality of the script to make it usable when (re)deploying a large number of instances.

## 💭 Motivation and Context ##

There was a mistake with one of the arguments being passed to Terraform when deploying the new instance that caused the Terraform step to fail. Additionally, this script previously parsed the Terraform state and terminated each instance sequentially. This led to potentially extremely long processing times when redeploying multiple instances.

This performance issue was resolved by parsing the Terraform state once using `jq`, and then checking the output of that for the instance ID of a given instance. This allows us to terminate and wait for successful termination of all desired instances at once. Consolidating these processes has greatly reduced the processing time for this script, making it usable for large redeployment tasks.

## 🧪 Testing ##

I have used this in my testing environment and @dav3r has used this script to redeploy an instance in the production environment.

## 🚥 Types of Changes ##

* Bug fix (non-breaking change which fixes an issue)
* New feature (non-breaking change which adds functionality)

## ✅ Checklist ##

* [x] PR has an informative and human-readable title
* [x] Changes are limited to a single goal - _eschew scope creep!_
* [x] Assignee and relevant labels have been added
* [x] I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document
* [x] Code changes follow [cisagov code standards](https://github.com/cisagov/development-guide)
* [x] All new and existing tests pass

### DIFF
--- a/terraform/deploy_new_portscan_instance.sh
+++ b/terraform/deploy_new_portscan_instance.sh
@@ -14,31 +14,48 @@ function usage {
     exit 1
 }
 
+function check_dependencies {
+    if [ -z "$(command -v terraform)" ] || \
+        [ -z "$(command -v aws)" ] || \
+        [ -z "$(command -v jq)" ]
+    then
+        echo "This script requires the following tools to run:"
+        echo "- terraform"
+        echo "- aws (AWS CLI)"
+        echo "- jq"
+        exit 1
+    fi
+}
+
 function redeploy_instances {
-    tf_args=("-var-file=\"$workspace.tfvar\"")
+    tf_args=()
+    portscanner_ids_json=$(terraform show -json | \
+        jq '.values.root_module.resources[] | select(.address == "aws_instance.cyhy_nmap" and .deposed_key == null) | {index, id: .values.id}' \
+        | jq -n '[inputs]')
+    nmap_instance_ids=()
+
     for index in $(seq "$1" "$2")
     do
         # Strip control characters, then look for the text "id" surrounded by
         # space characters, then extract only the ID from that line.
         # The first sed line has been carefully crafted to work with BSD sed.
-        nmap_instance_id=$(terraform state show aws_instance.cyhy_nmap[$index] | \
-                               sed $'s,\x1b\\[[0-9;]*[[:alpha:]],,g' | \
-                               grep "[[:space:]]id[[:space:]]" | \
-                               sed "s/[[:space:]]*id[[:space:]]*= \"\(.*\)\"/\1/")
+        nmap_instance_ids+=("$(echo "$portscanner_ids_json" | jq --raw-output ".[] | select(.index == $index) | .id")")
 
-        # Terminate the existing nmap instance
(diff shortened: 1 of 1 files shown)


dav3r (approved):
Very nice - well done! ✌️

terraform/deploy_new_portscan_instance.sh
```diff
@@ -14,31 +14,53 @@ function usage {
...
+
 function redeploy_instances {
-    tf_args=("-var-file=\"$workspace.tfvar\"")
+    tf_args=()
+    # Get all portscan instance IDs as a JSON array of dicts in the form:
+    # {
+    #   index: instance_index,
+    #   id: instance_id
+    # }
+    # Any previously removed instances are ignored (.deposed_key == null)
```
jsf9k, line 47: "previously removed" :arrow_right: "previously-removed"

terraform/deploy_new_portscan_instance.sh
```diff
@@ -14,31 +14,53 @@ function usage {
...
+    then
+        echo "This script requires the following tools to run:"
+        echo "- terraform"
+        echo "- aws (AWS CLI)"
+        echo "- jq"
+        exit 1
+    fi
+}
+
 function redeploy_instances {
```
jsf9k, line 40: I think a comment is needed here stating 
* What the arguments to this function are (`start` and `stop`)
* Whether the `start` to `stop` range is exclusive (`[start, stop)`) or inclusive (`[start, stop]`)

jsf9k (approved):
Powerful work!  Very nice improvement to this script!

I left two comments.  One is a miniscule grammar suggestion, which can be ignored if you disagree.  The other requests some additional documentation of the `redeploy_instances` function.

terraform/deploy_new_portscan_instance.sh
mcdonnnj: I added some comments and addressed your concern in both a comment about the function, and as a note in the usage output in 85aeda2. Thanks for pointing out that it was ambiguous 🐧

terraform/deploy_new_portscan_instance.sh
mcdonnnj: I *think* that since `previously` is an adverb I'm safe. Thank you for bringing up the suggestion though!

jsf9k: Ah, you are correct.  I have learned something.

terraform/deploy_new_portscan_instance.sh
jsf9k: What you wrote is perfect.  Thanks for clearing that up.

jsf9k (approved):
This looks great.  Thanks for clearing up the inclusive vs exclusive issue.

PR #331, opened 2021-06-30
Bump t3.micro Instance Usage to t3.small

mcdonnnj:
#  #

## 🗣 Description ##

This PR updates any configuration that uses the `t3.micro` instance type to use a `t3.small` instance instead.

I was unable to successfully build a `bastion` instance when preparing to test for a redeployment. The problem was a timeout when waiting for the ClamAV base definitions file to be created. I tested with the `nmap` type with the same result. Upon investigating I found log messages with `Can't allocate memory` during the initial update. After conferring with @jsf9k he agreed that bumping the instance type for more memory was the likely solution. After changing the build instance for the `bastion` type to a `t3.small` I was able to successfully create an AMI with Packer.

The running instance types are also updated to prevent this issue from arising again while running.

## 🧪 Testing ##

I was able to successfully build a `bastion` AMI once the change was made.

## ✅ Checklist ##

* [x] This PR has an informative and human-readable title.
* [x] Changes are limited to a single goal - _eschew scope creep!_
* [x] All relevant type-of-change labels have been added.
* [x] I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document.
* [x] These code changes follow [cisagov code standards](https://github.com/cisagov/development-guide).
* [x] All new and existing tests pass.

### DIFF
--- a/packer/bastion.json
+++ b/packer/bastion.json
@@ -16,7 +16,7 @@
         "us-west-1",
         "us-west-2"
       ],
-      "instance_type": "t3.micro",
+      "instance_type": "t3.small",
       "launch_block_device_mappings": [
         {
           "delete_on_termination": true,
--- a/packer/dashboard.json
+++ b/packer/dashboard.json
@@ -16,7 +16,7 @@
         "us-west-1",
         "us-west-2"
       ],
-      "instance_type": "t3.micro",
+      "instance_type": "t3.small",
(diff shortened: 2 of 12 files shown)


jsf9k (approved):
Great detective work!

dav3r (approved):
Nice job figuring this out 🔍

PR #336, opened 2021-09-08
Upgrade Terraform Version Used from 0.12 to 0.13

mcdonnnj:
#  #

## 🗣 Description ##

This PR updates the Terraform version that this project pins against from Terraform 0.12 to 0.13 and makes corresponding changes to the Terraform configuration.

## 💭 Motivation and context ##

This mirrors work in https://github.com/cisagov/cool-system/issues/94 to bring our entire org to Terraform 0.13 at the same time as bringing the COOL configurations (which are the vast majority of our Terraform code) to Terraform 0.13. The initial changes were done using `terraform 0.13upgrade` with a Terraform 0.13 binary. However, due to changes in Terraform 0.13 to bring full functionality to modules this PR also had to dismantle the previous approach of dynamic modules for Nessus and nmap instances and makes them regular modules in the Terraform configuration.

## ⚙ Process ##

I first worked on a clean `terraform0.12 apply` with the existing Terraform 0.12 state for each Terraform configuration in this repository (`terraform_egress_pub/`, `terraform_route_53/`, `terraform_nessus_only/`, `terraform/`) using the `develop` branch. Once that was ensured I switched to this branch, ran `terraform0.13 init -reconfigure`, and then attempted a `terraform0.13 apply`. If there were any issues I resolved them and proceeded with the `apply` until it was clean.

### `terraform_egress_pub/` ###

This had no issues with the existing Terraform 0.12 state and was fine for a Terraform 0.13 `apply` with the existing code.

### `terraform_route_53/` ###

When attempting to perform a `terraform0.12 apply` the resource

https://github.com/cisagov/cyhy_amis/blob/f4551046bb205dc5dedf517745af45731001ca97/terraform_route_53/cyber.dhs.gov.tf#L122-L128

needed to be created. This should exist as part of the configuration described in [ciagov/cool-dns-cyber.dhs.gov](https://github.com/cisagov/cool-dns-cyber.dhs.gov/blob/6a4050b4480f1985fa20623f836bff35dec72783/route53_dmarc_app.tf#L38-L46), but for some reason did not. After a team discussion we decided to let it get created to move forward. Once this apply was completed there were no issues with a successful Terraform 0.13 `apply`.

### `terraform_nessus_only/` ###

We have determined that this configuration will be completely spun-down and something equivalent will be spun up in the COOL. This configuration **remains** Terraform 0.12 and will have a `terraform0.12 destroy` once the migration occurs as opposed to updating the state to Terraform 0.13.

### `terraform/` ###

[...]

PR #355, opened 2022-02-11
Use the `default_tags` provider argument

### LINKED ISSUE #327: Use `default_tags` in Providers
# 🚀 Feature Proposal

We should switch to using the `default_tags` block in providers to manage common tags. The version of the AWS provider this requires means this relies on #326 being resolved before it can be implemented.

## Motivation

This would follow what was done in https://github.com/cisagov/skeleton-tf-module/pull/49 and is implemented (or in the process of being implemented) for other Terraform projects now.

## Pitch

Although this project descends from [cisagov/skeleton-generic](https://github.com/cisagov/skeleton-generic) we should still try to incorporate changes made in [cisagov/skeleton-tf-module](https://github.com/cisagov/skeleton-tf-module) and [cisagov/skeleton-packer](https://github.com/cisagov/skeleton-packer) whenever possible.

mcdonnnj:
#  #

## 🗣 Description ##

This pull request updates the Terraform configurations to use the `default_tags` argument instead of per-resource `tag` arguments. This closes #327.

There is also a touch of configuration modernization such as renaming `terraform.tf` files to `backend.tf` and removing all `main.tf` files in favor of putting the contents in the `locals.tf` and `providers.tf` files.

Once https://github.com/cisagov/cyhy_amis/pull/336 was merged the AWS provider version in use supported the `default_tags` argument. Using the `default_tags` argument makes resource tag management much easier and ensures that all resources that _can_ be tagged will at least received the base tags given in the `tags` variable.

## 🧪 Testing ##

Automated tests pass. Performing a `terraform plan` shows tag changes (in additional to Ansible provisionr signature changes, but that's not unexpected).

- [x] Changes are limited to a single goal - _eschew scope creep!_


terraform/locals.tf
```diff
@@ -1,3 +1,19 @@
...
+
+# ------------------------------------------------------------------------------
+# Retrieve the default tags for the default provider.  These are
+# used to create volume tags for EC2 instances, since volume_tags does
+# not yet inherit the default tags from the provider.  See
+# hashicorp/terraform-provider-aws#19188 for more details.
+# ------------------------------------------------------------------------------
+data "aws_default_tags" "default" {
+  provider = aws
+}
```
jsf9k: Why do we do this instead of referencing `var.tags` explicitly?

jsf9k (commented):
I have one question.

terraform/locals.tf
mcdonnnj: That's what I saw in [other projects](https://github.com/search?q=org%3Acisagov+aws_default_tags&type=code) and I pulled it from the skeleton [here](https://github.com/cisagov/skeleton-tf-module/blob/f1c81f4e0ec92f988c76def81d26786f5688d086/main.tf#L37-L38) and [here](https://github.com/cisagov/skeleton-tf-module/blob/f1c81f4e0ec92f988c76def81d26786f5688d086/main.tf#L55-L60).

mcdonnnj: This could be a misinterpretation on my part because those projects often have multiple providers, but I figured I would follow precedent.

mcdonnnj: In [a2de9a8](https://github.com/cisagov/cyhy_amis/pull/355/commits/a2de9a86237e768dbb7ab25d9de764e8281d850b) I removed the `provider` to keep in line with the examples I linked.

jsf9k (approved):

dav3r (approved):
Looks solid to me. 👍

PR #408, opened 2022-04-19
Increase Nessus production root disk size to 200 GB

dav3r:
#  #

## 🗣 Description ##

This PR increases the Nessus production root disk size to 200 GB.

More disk means more uptime without running out of space.  We just had to deal with an instance that ran out of disk space and it's something we'd prefer to avoid.

This change has already been applied in Production.

## 🧪 Testing ##

I ran a targeted apply in Production and verified that everything worked as intended.  Note that I had to follow [steps similar to these](https://github.com/cisagov/cyhy-system/wiki/Adjust-reporter-partition-and-filesystem-size) in order to get the instances to recognize the larger disks.


mcdonnnj (approved):
:shipit:

jsf9k (approved):

PR #394, opened 2022-04-01
Add CloudWatch alarm notifications for certain failed data ingestion processes

jsf9k:
## 🗣 Description ##

This pull request adds CloudWatch alarms and SNS topics for failed NVD and KEV data ingestion processes.

## 💭 Motivation and context ##

Contributes to the resolution of cisagov/cyhy-system#37.

See also:
- cisagov/cyhy-core#67
- cisagov/cyhy_amis#396
- cisagov/cyhy_amis#397

## 🧪 Testing ##

None yet, other than my ocular orbs.

## ✅ Pre-approval checklist ##

- [x] This PR has an informative and human-readable title.
- [x] Changes are limited to a single goal - *eschew scope creep!*
- [x] All relevant type-of-change labels have been added.
- [x] I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document.
- [x] These code changes follow [cisagov code standards](https://github.com/cisagov/development-guide).
- [x] All relevant repo and/or project documentation has been updated to reflect the changes in this PR.
- [x] All new and existing tests pass.

PR #424, opened 2022-04-28
Update the cloud-init for setting instance hostnames

mcdonnnj:
#  #

## 🗣 Description ##

This pull request changes from using a bash script to set instance hostnames to using the [cloud-init] [set_hostname module](https://cloudinit.readthedocs.io/en/latest/topics/modules.html?highlight=hostname#set-hostname). As part of that the [cloud-init] configurations are overhauled to give each instance its own `*_cloud_init.tf` file. Lastly filenames are updated for [cloud-init] tasks that are not set up for templating.

This change is necessary to support the work in #394 because the instance's hostname is based on the private IP address the instance receives by default. This change ensures that a value we expect is used for the hostname and is correctly seen by the CloudWatch Agent.

## 🧪 Testing ##

Automated tests pass. I deployed this in my testing environment and verified that instances that are currently able to push logs (there are network restrictions that impact some of the instance types currently) uploaded to the appropriate `/instance-logs/<instance-type>` log group. Specifically in the case of the alarms the database was pushing its logs to `/instance-logs/database1` as expected.

### ⚠ Caveats ###

The private domain zone name is **not** included in the `local_hostname` value (on Debian at least). This means that for the bastion instances that see both a `cyhy` and a `bod` deployment they would both upload logs to the `/instance-logs/bastion` group. This also means that multiple deployments in the same region will push to the same log groups, but this should not impact the production environment given how we deploy things.

- [x] All relevant repo and/or project documentation has been updated
      to reflect the changes in this PR.

[cloud-init]: https://cloud-init.io/


jsf9k (approved):
Looks great to me!

dav3r (approved):
Excellent work! 👍

PR #427, opened 2022-05-01
Fix EBS volume destruction

mcdonnnj:
#  #

## 🗣 Description ##

This pull request does three things:
1. Switch from the `skip_destroy` argument to the `stop_instance_before_detaching` argument for `aws_volume_attachment` resources.
1. Change to using the `*_instance_count` variables for the `count` argument in the `nmap` and `nessus` related resources.
1. Make each instance type depend on any EBS volumes it uses.

Changing the argument used is to help ensure volume detachment proceeds smoothly. We can swap them like this only because the volume attachments we use are designed to be for the life of the instance. Removing the explicit dependency of using `length(aws_instance.<instance_type>)` not only allows the EBS volumes that do not have any implicit dependencies to be created in parallel, but also allows the instances to be destroyed in the same way. Without this change the instances will not be destroyed until after the EBS volumes are destroyed, but the EBS volumes cannot be destroyed while the instances are live. Lastly since the EBS volumes used by each instance are necessary for regular functions we make each instance `depends_on` the EBS volumes it uses. This should guarantee they are available for attachment as soon as the respective instance is created.

## 🧪 Testing ##

Automated tests pass. I was able to successfully tear down an existing deployment, perform a greenfield apply, and then destroy that deployment.

- [x] All relevant repo and/or project documentation has been updated
      to reflect the changes in this PR.

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -146,15 +146,15 @@ terraform apply -var-file=<your_workspace>.tfvars
 | Name | Version |
 |------|---------|
 | terraform | ~> 1.0 |
-| aws | ~> 3.38 |
+| aws | ~> 3.62 |
 | cloudinit | ~> 2.0 |
 
 ## Providers ##
 
 | Name | Version |
 |------|---------|
-| aws | ~> 3.38 |
-| aws.public\_dns | ~> 3.38 |
+| aws | ~> 3.62 |
+| aws.public\_dns | ~> 3.62 |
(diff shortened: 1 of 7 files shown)


jsf9k (approved):
This is a great improvement!  @dav3r, we should adopt this method for any EBS disk attachments we have in the COOL.

dav3r (approved):
Outstanding! 🚀

PR #429, opened 2022-05-01
Ensure `mongo` instance cloud-init scripts run in the necessary order

mcdonnnj:
#  #

## 🗣 Description ##

This pull request updates the local names for the `cloud-init` shellscripts to ensure that they run in the desired order.

In #424 I refactored the `cloud-init` configurations for each of the instances, but in the case of the `mongo` instances I neglected to account for the fact that `cloud-init` runs them in the order given by the Python `sorted()` function. This resulted in the journal shellscript running before the data shellscript, which is a problem because the journal is in a subdirectory of the data directory. As a result `cloud-init` failed while running and MongoDB could not successfully start.

## 🧪 Testing ##

Automated tests pass. I ensured that redeploying with this change in place resulted in the `mongod` service successfully starting instead of failing a `RequiresMountsFor` check.

### DIFF
--- a/terraform/cyhy_mongo_cloud_init.tf
+++ b/terraform/cyhy_mongo_cloud_init.tf
@@ -24,13 +24,13 @@ data "cloudinit_config" "cyhy_mongo_cloud_init_tasks" {
       num_disks     = 4
     })
     content_type = "text/x-shellscript"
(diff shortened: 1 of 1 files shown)


jsf9k (approved):

dav3r (approved):
Good catch and fix! 🐞

PR #435, opened 2022-05-03
Increase volume sizes on reporter and vulnscan instances

dav3r:
#  #

## 🗣 Description ##

This PR increases the size of two volumes:
* The data volume for Production reporter instances
* The root volume for non-Production vulnscan instances

The Production reporter volume was almost 90% full, so it made sense to expand it since we are now adding about 5 GB of reports each week.

The non-Production vulnscan root volume needed to be expanded because it kept filling up during my most recent round of testing.  This is likely a combination of the ever-growing number of vulnerability plugins that must be downloaded, as well as the general hunger for more disk space as newer operating system software is added.

## 🧪 Testing ##

The reporter change has been applied in Production and verified.
The vulnscan change has been applied in a non-Production environment and verified.

### DIFF
--- a/terraform/cyhy_nessus_ec2.tf
+++ b/terraform/cyhy_nessus_ec2.tf
@@ -29,7 +29,7 @@ resource "aws_instance" "cyhy_nessus" {
   subnet_id = aws_subnet.cyhy_vulnscanner_subnet.id
 
   root_block_device {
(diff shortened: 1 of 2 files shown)


mcdonnnj (approved):
:shipit:

jsf9k (approved):

PR #436, opened 2022-05-04
Ensure filesystems are mounted before performing a `chown` on them

jsf9k:
## 🗣 Description ##

This pull request moves the `chown`ing of directories from the second Ansible run to `cloud-init`.  If the directory to be `chown`ed happens to be a mount point, then this code ensures that it is mounted before performing the `chown`.

When redeploying CyHy recently, @dav3r noticed that the ownership of files in some mounted filesystems was incorrect.

## 🧪 Testing ##

I created a testing deployment of CyHy with these changes and verified (on the instance types that I could `ssh` into) that it functions as expected.


ansible/roles/cyhy_dashboard/meta/main.yml
```diff
@@ -52,7 +52,6 @@ galaxy_info:
     # NOTE: A tag is limited to a single word comprised of alphanumeric characters.
     #       Maximum 20 tags per role.
 
-dependencies:
+dependencies: []
   # List your role dependencies here, one per line. Be sure to remove the '[]' above,
   # if you add dependencies to this list.
```
dav3r: Might as well remove these comments here, since they are not present in the other `meta/main.yml` files.

terraform/bod_docker_cloud_init.tf
```diff
@@ -38,6 +38,39 @@ data "cloudinit_config" "bod_docker_cloud_init_tasks" {
     filename     = "vdp_disk_setup.sh"
   }
 
+  part {
+    content = templatefile("${path.module}/cloud-init/chown_directory.tpl.sh", {
+      group          = "cyhy"
+      is_mount_point = true
+      owner          = "cyhy"
+      path           = "/var/cyhy/orchestrator/output"
```
dav3r, line 45: Do we want to replace `/var/cyhy` with a local variable since it's now repeated a bunch of times in our Terraform?  I'm fine either way- just throwing it out there.

dav3r (approved):
This all looks fine to my eyes- thanks for takin' care of business! 👍

terraform/bod_docker_cloud_init.tf
mcdonnnj: Unless we change the Ansible roles creating some of these directories I'd rather leave it hard-coded to make it clear that it's not expected to change.

ansible/roles/cyhy_dashboard/meta/main.yml
jsf9k: Please see commit 8849ee1.

terraform/bod_docker_cloud_init.tf
jsf9k: I agree with @mcdonnnj, and I like it being made explicit in the `*_cloud_init.tf` file what directories are being `chown`ed.  With the local variable I'd have to go to another file to look up the value when reading the Terraform code.

dav3r: Those are good reasons in my book. 👍

terraform/bod_docker_cloud_init.tf
```diff
@@ -38,6 +38,39 @@ data "cloudinit_config" "bod_docker_cloud_init_tasks" {
...
+      group          = "cyhy"
+      is_mount_point = true
+      owner          = "cyhy"
+      path           = "/var/cyhy/vdp/output"
+    })
+    content_type = "text/x-shellscript"
+    filename     = "cyhy_docker_chown_vdp_output_directory.sh"
+  }
+
+  part {
```
mcdonnnj, line 62: Would you re-arrange these `part`s so they are ordered how we expect them to run on the instance?

terraform/cyhy_nessus_cloud_init.tf
```diff
@@ -27,6 +27,28 @@ data "cloudinit_config" "cyhy_nessus_cloud_init_tasks" {
...
+      group          = "cyhy"
+      is_mount_point = true
+      owner          = "cyhy"
+      path           = "/var/cyhy/runner"
+    })
+    content_type = "text/x-shellscript"
+    filename     = "cyhy_nessus_chown_runner_directory.sh"
+  }
+
+  part {
```
mcdonnnj, line 53: Same ask regarding `part` ordering.

terraform/cyhy_nmap_cloud_init.tf
```diff
@@ -27,6 +27,28 @@ data "cloudinit_config" "cyhy_nmap_cloud_init_tasks" {
...
+      group          = "cyhy"
+      is_mount_point = true
+      owner          = "cyhy"
+      path           = "/var/cyhy/runner"
+    })
+    content_type = "text/x-shellscript"
+    filename     = "cyhy_nmap_chown_runner_directory.sh"
+  }
+
+  part {
```
mcdonnnj, line 53: Same ask regarding `part` ordering.

terraform/cyhy_reporter_cloud_init.tf
```diff
@@ -25,6 +25,39 @@ data "cloudinit_config" "cyhy_reporter_cloud_init_tasks" {
     filename     = "cyhy_reports_disk_setup.sh"
   }
 
+  part {
```
mcdonnnj, line 40: Same ask regarding `part` ordering.

mcdonnnj (approved):
Thanks for taking care of this! I do have a set of requests around ordering the `part`s to correspond with the order they are expected to run on the instance.

terraform/bod_docker_cloud_init.tf
jsf9k: I can certainly reorder the script `part`s in that order.  I believe the order in which the various `text/cloud-config` bits run is dependent on the contents and has nothing to do with the `part`s.  One such `part` could contain two pieces of content that are run at different stages, but for the most part they will all run _before_ the user scripts.

Shall I just put all the `text/cloud-config` pieces at the top of the file (ordered by their `filename` attribute) and all the `text/x-shellscript` parts below that (ordered according to how they are expected to run on the instance)?

mcdonnnj: Yes you're right about `text/cloud-config` ordering and I'm totally fine with your solution as that's what I was thinking myself.

terraform/cyhy_reporter_cloud_init.tf
jsf9k: Please see commit ddfa965.

terraform/cyhy_nmap_cloud_init.tf
jsf9k: Please see commit ddfa965.

terraform/cyhy_nessus_cloud_init.tf
jsf9k: Please see commit ddfa965.

terraform/bod_docker_cloud_init.tf
jsf9k: Please see commit ddfa965.

terraform/bod_docker_cloud_init.tf
```diff
@@ -22,7 +34,29 @@ data "cloudinit_config" "bod_docker_cloud_init_tasks" {
...
+    content = templatefile("${path.module}/cloud-init/chown_directory.tpl.sh", {
+      group          = "cyhy"
+      is_mount_point = false
+      owner          = "cyhy"
+      path           = "/var/cyhy"
+    })
+    content_type = "text/x-shellscript"
+    filename     = "02_cyhy_docker_chown_cyhy_directory.sh"
```
mcdonnnj, line 59: This used to run before the `cyhy_docker_chown_orchestrator_output_directory.sh` script because of the default sorting from cloud-init (Python `sorted()` function). Are we ok to change this ordering?

terraform/cyhy_nessus_cloud_init.tf
```diff
@@ -24,18 +36,28 @@ data "cloudinit_config" "cyhy_nessus_cloud_init_tasks" {
...
+    content = templatefile("${path.module}/cloud-init/chown_directory.tpl.sh", {
+      group          = "cyhy"
+      is_mount_point = false
+      owner          = "cyhy"
+      path           = "/var/cyhy"
+    })
+    content_type = "text/x-shellscript"
+    filename     = "02_cyhy_nessus_chown_cyhy_directory.sh"
```
mcdonnnj, line 61: Same question about changed ordering.

terraform/cyhy_nmap_cloud_init.tf
```diff
@@ -24,18 +36,28 @@ data "cloudinit_config" "cyhy_nmap_cloud_init_tasks" {
...
+    content = templatefile("${path.module}/cloud-init/chown_directory.tpl.sh", {
+      group          = "cyhy"
+      is_mount_point = false
+      owner          = "cyhy"
+      path           = "/var/cyhy"
+    })
+    content_type = "text/x-shellscript"
+    filename     = "02_cyhy_nmap_chown_cyhy_directory.sh"
```
mcdonnnj, line 61: Same question about changed ordering.

terraform/cyhy_reporter_cloud_init.tf
```diff
@@ -22,18 +34,39 @@ data "cloudinit_config" "cyhy_reporter_cloud_init_tasks" {
...
+
+  part {
+    content = templatefile("${path.module}/cloud-init/chown_directory.tpl.sh", {
+      group          = "cyhy"
+      is_mount_point = false
+      owner          = "cyhy"
+      path           = "/var/cyhy"
+    })
+    content_type = "text/x-shellscript"
+    filename     = "02_cyhy_reports_chown_cyhy_directory.sh"
```
mcdonnnj, line 59: Same question about ordering.

mcdonnnj (commented):
I had some questions about how your adjusted `filename` values would change the existing execution order.

terraform/bod_docker_cloud_init.tf
jsf9k: I think we are.  We could probably get away with only `chown`ing the parent directory, but it doesn't hurt to do the child directory (and mount point) first.

The disk is already set up when the script runs, and the script waits until the mount point is available before starting.

terraform/cyhy_nessus_cloud_init.tf
jsf9k: I think we are.  We could probably get away with only `chown`ing the parent directory, but it doesn't hurt to do the child directory (and mount point) first.

The disk is already set up when the script runs, and the script waits until the mount point is available before starting.

terraform/cyhy_nmap_cloud_init.tf
jsf9k: I think we are.  We could probably get away with only `chown`ing the parent directory, but it doesn't hurt to do the child directory (and mount point) first.

The disk is already set up when the script runs, and the script waits until the mount point is available before starting.

terraform/cyhy_reporter_cloud_init.tf
jsf9k: I think we are.  We could probably get away with only `chown`ing the parent directory, but it doesn't hurt to do the child directory (and mount point) first.

The disk is already set up when the script runs, and the script waits until the mount point is available before starting.

PR #441, opened 2022-05-13
Update the Ansible provisioner settings in our Packer configuration

mcdonnnj:
#  #

## 🗣 Description ##

This pull request updates the Ansible provisioner configuration in the following ways:
- Set `use_proxy` to `false` so that the provisioner will only create a local proxy adapter if it cannot determine an IP address for the target instance. As part of that change the `ansible.cfg` file is deleted.
- Force the provisioner to use SFTP to transfer files to the target instance.
- Turn the `build_region` and `ami_regions` values into variables with defaults.
- Set the default region for AWS.

The biggest change is to set `use_proxy` to `false` and this is made to resolve issues with newer version of OpenSSH on the machine initiating the Packer build. Without the local proxy adapter we no longer need to adjust the Ansible configuration to tell the local SSH configuration to allow RSA keys. The root issue is that the local proxy adapter is set up with an RSA key (we cannot change this at present) which runs afoul of OpenSSH 8.8+ as these versions disable support for RSA keys by default.

We force SFTP to mirror https://github.com/cisagov/skeleton-packer/pull/136. Switching `build_region` and `ami_regions` to variables is just another piece of trying to organize these older configurations more like our modern configurations descended from [cisagov/skeleton-packer](https://github.com/cisagov/skeleton-packer). Lastly we set `AWS_DEFAULT_REGION` for the provisioner so that we are not hard-coding regions in things like SSM lookups.

This blurb from @jsf9k sums up two of the changes nicely:

> It's worth noting that `use_proxy=false` and `use_sftp=true` force Packer's Ansible provisioner to use Ansible as it functions by default; the default settings for Packer's Ansible provisioner are actually _overriding_ Ansible's default behavior.

## 🧪 Testing ##

Automated tests pass. I pulled these changes into my testing branch and was able to build AMIs with OpenSSH 8.9p1 and then OpenSSH 9.0 (current release) with no issues.


dav3r (approved):
Excellent work- this is beautiful 🌷

jsf9k (approved):
It's worth noting that `use_proxy=false` and `use_sftp=true` force Packer's Ansible provisioner to use Ansible as it functions by default; the default settings for Packer's Ansible provisioner are actually _overriding_ Ansible's default behavior.

PR #460, opened 2022-06-06
Update the version of Nessus installed on `nessus` AMIs

mcdonnnj:
#  #

## 🗣 Description ##

This pull request explicitly states and updates the version of Nessus installed on `nessus` AMI types. It also changes the arguments to [cisagov/ansible-role-nessus] to use a block-style map instead of flow-style.

We haven't updated the version of Nessus installed in quite some time and we need a minimum version of `8.14` to get Vulnerability Priority Rating (VPR) support. Since we need to update the version and instead of relying on an obfuscated software version installed as a default for the Ansible role we're using it makes sense to explicitly set the version we want used in `nessus` AMIs. 

## 🧪 Testing ##

Automated tests pass. I verified that when deployed as a `vulnscanner` instance scanning ran without issue.

[cisagov/ansible-role-nessus]: https://github.com/cisagov/ansible-role-nessus

### DIFF
--- a/packer/ansible/playbook.yml
+++ b/packer/ansible/playbook.yml
@@ -45,7 +45,7 @@
   become_method: sudo
   roles:
     - cyhy_runner
-    - {role: nessus, package_bucket: ncats-3rd-party-packages}
+    - {role: nessus, package_bucket: ncats-3rd-party-packages, version: "8.15.5"}
     - more_ephemeral_ports
 
 - hosts: bod
--- a/packer/ansible/playbook.yml
+++ b/packer/ansible/playbook.yml
@@ -45,9 +45,10 @@
   become_method: sudo
   roles:
     - cyhy_runner
-    - package_bucket: ncats-3rd-party-packages
-      role: nessus
-      version: "8.15.5"
+    - role: nessus
+      vars:
+        package_bucket: ncats-3rd-party-packages
+        version: "8.15.5"
     - more_ephemeral_ports
 
 - hosts: bod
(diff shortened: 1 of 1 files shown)


packer/ansible/playbook.yml
```diff
@@ -45,7 +45,9 @@
   become_method: sudo
   roles:
     - cyhy_runner
-    - {role: nessus, package_bucket: ncats-3rd-party-packages}
+    - package_bucket: ncats-3rd-party-packages
+      role: nessus
+      version: "8.15.5"
```
jsf9k: FYI, this is the default version as of cisagov/ansible-role-nessus#16.

jsf9k (approved):

packer/ansible/playbook.yml
dav3r: @mcdonnnj alluded to this in this PR's "Motivation and context" section above:
> Since we need to update the version and instead of relying on an obfuscated software version installed as a default for the Ansible role we're using it makes sense to explicitly set the version we want used in nessus AMIs.

dav3r (approved):
Yup yup yup! ⚙️

packer/ansible/playbook.yml
mcdonnnj: This also allows the Ansible role to use whatever version it wants as a default while allowing the CyHy images to remain on a version tested against our orchestration.

jsf9k: I wasn't suggesting changing it; I was just pointing it out in case anyone cared.  _Conversation resolved!_

jsf9k (approved):

PR #471, opened 2022-06-21
Add 4 additional vulnscan instances in `commander.conf`

dav3r:
#  #

## 🗣 Description ##

This PR adds 4 additional vulnscan instances to `commander.conf`.

## 💭 Motivation and context ##

This change supports our increased scanning workload.

## 🧪 Testing ##

This change will be verified by manually editing this file in Production after the new vulnscan instances have been deployed.

## ✅ Pre-approval checklist ##

- [x] This PR has an informative and human-readable title.
- [x] Changes are limited to a single goal - *eschew scope creep!*
- [x] All relevant type-of-change labels have been added.
- [x] I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document.
- [x] These code changes follow [cisagov code standards](https://github.com/cisagov/development-guide).
- [x] All new and existing tests pass.

## ✅ Pre-merge checklist ##

- [x] Testing in Production has been confirmed successfully.

PR #477, opened 2022-06-27
Reduce to 6 vulnscan instances in `commander.conf`

dav3r:
#  #

## 🗣 Description ##

This PR removes two vulnscan instances in `commander.conf`, bringing us to six.

When we ramped up to eight instances in #471, that turned out to be more than we needed.

Note that this also corrects an error from #471, where the only the "purge-production" section was updated, and not the "production" section that we primarily use.

## 🧪 Testing ##

This change was verified by manually editing this file in Production and restarting the commander after the 2 unneeded vulnscan instances were destroyed.


mcdonnnj (approved):
:shipit:

felddy (approved):
Epic!
⛵ 🇬🇷

PR #478, opened 2022-07-01
Modernize the Ansible roles used in the Terraform configuration

mcdonnnj:
#  #

## 🗣 Description ##

This pull request modernizes the Ansible roles we use in our Terraform configuration. This modernization updates them to more closely resemble our `cisagov/ansible-role-*` roles.

## 💭 Motivation and context ##

We currently do not perform any linting of these Ansible roles. Modernizing them not only makes sure they are written similarly to our `cisagov/ansible-role-*` roles but also allows us to enable the `ansible-lint` hook for `pre-commit`.

## 🧪 Testing ##

Automated tests pass. I used this branch in my [testing branch](https://github.com/cisagov/cyhy_amis/tree/testing/redeployment_updates) and was able to successfully deploy my instances. I checked the files on the production systems and in my testing environment and I am only seeing expected changes due to formatting, different regions, etc.

## ✅ Pre-approval checklist ##

- [x] This PR has an informative and human-readable title.
- [x] Changes are limited to a single goal - *eschew scope creep!*
- [x] All relevant type-of-change labels have been added.
- [x] I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document.
- [x] These code changes follow [cisagov code standards](https://github.com/cisagov/development-guide).
- [x] Tests have been added and/or modified to cover the changes in this PR.
- [x] All new and existing tests pass.

PR #542, opened 2022-10-19
Use a file to store the configuration to access the Nessus API

mcdonnnj:
#  #

## 🗣 Description ##

This pull request creates a file that contains information for interacting with Nessus on any instances that run the `nessus` Ansible role defined in this repository (the `vulnscan` instances in our Terraform configuration). The Python script that does some post-deployment configuration of Nessus is updated to use this file instead of having information templated into it by the Ansible role.

We are looking to make changes to jobs that are pushed to `vulnscan` instances that require the configuration information to be available on the instance somewhere. Since we need this information available it makes sense to store them in a static location on the instance and to update the configuration Python script to use this file.

## 🧪 Testing ##

Automated tests pass. I used this in my testing environment and verified that Nessus was configured as it should be post deployment and that the configuration file was on the instance.

- [x] Tests have been added and/or modified to cover the changes in this PR.

### DIFF
--- a/ansible/roles/nessus/tasks/main.yml
+++ b/ansible/roles/nessus/tasks/main.yml
@@ -78,6 +78,22 @@
         state: started
   when: registered.rc != 0 or not matching_key|default(false)|bool
 
+- name: Create /etc/cyhy directory
+  file:
+    path: /etc/cyhy
+    state: directory
+    owner: cyhy
+    group: cyhy
+    mode: 0750
+
+- name: Create the configuration file for Nessus API access
+  template:
+    src: nessus_api.yml.j2
+    dest: /etc/cyhy/nessus_api.yml
+    owner: cyhy
+    group: cyhy
+    mode: 0640
+
 # Copy the nessus_base.py to tmp
 - name: Copy the nessus_base python file for importing the base nessus policy
   copy:
--- a/ansible/roles/nessus/tasks/main.yml
+++ b/ansible/roles/nessus/tasks/main.yml
@@ -94,14 +94,13 @@
     group: cyhy
     mode: 0640
 
-# Copy the nessus_base.py to tmp
-- name: Copy the nessus_base python file for importing the base nessus policy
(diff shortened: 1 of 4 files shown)


jsf9k (approved):
Nice improvement!

ansible/roles/nessus/tasks/main.yml
```diff
@@ -78,28 +78,29 @@
...
+    group: cyhy
+    mode: 0750
+
+- name: Create the configuration file for Nessus API access
+  template:
+    src: nessus_api.yml.j2
+    dest: /etc/cyhy/nessus_api.yml
+    owner: cyhy
+    group: cyhy
+    mode: 0640
```
dav3r, line 95: Alphabetize?
```suggestion
  file:
    group: cyhy
    mode: 0750
    owner: cyhy
    path: /etc/cyhy
    state: directory

- name: Create the configuration file for Nessus API access
  template:
    dest: /etc/cyhy/nessus_api.yml
    group: cyhy
    mode: 0640
    owner: cyhy
    src: nessus_api.yml.j2
```

dav3r (approved):
Outstanding!  🥇 
I have but one minor quibble.

ansible/roles/nessus/tasks/main.yml
mcdonnnj: I intentionally did not alphabetize to match the existing Ansible role structures and I was planning to update #478 once this PR is merged which would have those changes.

dav3r: OK, I suspected you had some motive here, but I wasn't sure what it was.  Thanks for the explanation. 👍

mcdonnnj: Just to circle back I have rebased #478 and the concerns raised here should be resolved in https://github.com/cisagov/cyhy_amis/pull/478/commits/984084dc5dc8056eda10b189ccf25ffd42b54141.

felddy (commented):
Looks good!

PR #543, opened 2022-10-19
Use the next size up instance type for the `reporter` instance

mcdonnnj:
#  #

## 🗣 Description ##

This pull request changes the instance type for production `reporter` instances from `c5.4xlarge` to `c5.9xlarge`.

We have seen a specific organization fail to generate weekly reports due to running out of memory on the instance. This is due in part to the way our report generation process occurs (initial memory + fork means that using close to half the instance memory will cause a report to fail to generate when it attempts to fork). We bumped the instance type in production manually and have not seen report generation failures since.

## 🧪 Testing ##

Automated tests pass. This change was made in production over a month ago and no problems have arisen.


jsf9k (approved):

dav3r (approved):
👍 👍

PR #550, opened 2022-10-24
Increase texmf buffer size for reporter

jsf9k:
## 🗣 Description ##

This pull request increases the `texmf` buffer size for the CyHy reporter instance.

TeX uses the buffer to contain input lines, but macro expansion works by writing material into the buffer and re-parsing the line. As a consequence, certain constructs require the buffer to be very large. As distributed by the OS package, the size is 200,000, which is more than sufficient for most documents; however, we have recently seen a lengthy CyHy report that required a value greater than 500,000. This is why this change is necessary.

## 🧪 Testing ##

All automated tests pass.

## ✅ Pre-merge checklist ##

- [ ] Revert dependencies to default branches.

### DIFF
--- a/packer/ansible/cyhy_reporter.yml
+++ b/packer/ansible/cyhy_reporter.yml
@@ -7,5 +7,7 @@
     - xfs
     - cyhy_reports
     - cyhy_mailer
+  vars:
+    texmf_buffer_size: 1000000
   vars_files:
     - vars/maxmind_license_key.yml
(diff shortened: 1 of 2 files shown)


dav3r (approved):
👍 👍

mcdonnnj (approved):
Using this pull request in my development environment's configuration I confirmed that the new `/etc/texmf/texmf.d/99buffer_size.cnf` was created and contains the value specified in this PR instead of the default in [cisagov/ansible-role-cyhy-reports]. I also confirmed that this was propagated to the main `textmf.cnf` file (with attribution for the file created by [cisagov/ansible-role-cyhy-reports]).

[cisagov/ansible-role-cyhy-reports]: https://github.com/cisagov/ansible-role-cyhy-reports

PR #559, opened 2022-11-08
Explicitly set the id for the `cyhy` user and group

mcdonnnj:
#  #

## 🗣 Description ##

This pull request updates the [cloud-init] configuration that sets up the `cyhy` user (and implicitly sets up the `cyhy` group) to use an explicitly defined id number. 

This is necessary to allow Docker images that use secrets with permissions restricted to the `cyhy` user to use a known UID in their configuration so that they are able to access those secrets when they are mounted into a running container. This also establishes known values for the `cyhy` uid/gid much like we do for some of the instance user accounts used in the COOL.

## 🧪 Testing ##

Automated tests pass. I confirmed after redeploying my test environment that the `cyhy` user has the expected uid on all instances that use this [cloud-init] configuration.

[cloud-init]: https://cloud-init.io/

### DIFF
--- a/terraform/cloud-init/cyhy_user_ssh_setup.yml
+++ b/terraform/cloud-init/cyhy_user_ssh_setup.yml
@@ -1,8 +1,9 @@
-#cloud-config
+---
 
 users:
-  - name: cyhy
+  - homedir: /var/cyhy
+    name: cyhy
     shell: /bin/bash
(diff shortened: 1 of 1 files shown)


dav3r (approved):
2️⃣ 0️⃣ 4️⃣ 8️⃣

jsf9k (approved):

PR #594, opened 2023-01-06
Add an additional resource to the `adi` and `fdi` Lambda deployment scripts

mcdonnnj:
#  #

## 🗣 Description ##

This pull request adds the respective `aws_iam_role_policy.<adi|fdi>_lambda_cloudwatch_policy` resources to the deployment scripts for the [adi] and [fdi] Lambdas.

If the Lambda name is changed and then redeployed using the corresponding deployment script the document that control CloudWatch permissions for logging is not updated without the corresponding resource. Adding these resources helps ensure that redeploying one of these Lambdas covers all of the resources required for the Lambda to function as expected.

## 🧪 Testing ##

Automated tests pass. I manually ran a `terraform apply` targeting these resources and verified that the [fdi] Lambda was able to generate log streams after the apply (it was unable to before the apply).

[adi]: https://github.com/cisagov/assessment-data-import-lambda
[fdi]: https://github.com/cisagov/findings-data-import-lambda


terraform/scripts/deploy_new_adi_lambda.sh
```diff
@@ -23,4 +23,5 @@ terraform apply -var-file="$workspace.tfvars" \
   -target=aws_lambda_function.adi_lambda \
   -target=aws_lambda_permission.adi_lambda_allow_bucket \
   -target=aws_s3_bucket_notification.adi_lambda \
-  -target=aws_cloudwatch_log_group.adi_lambda_logs
+  -target=aws_cloudwatch_log_group.adi_lambda_logs \
+  -target=aws_iam_role_policy.adi_lambda_cloudwatch_policy
```
dav3r: We should probably sort these targets alphabetically.

dav3r (approved):
👍  with one suggestion that applies to both scripts.

jsf9k (approved):
Approved, once @dav3r's concern is addressed.

terraform/scripts/deploy_new_adi_lambda.sh
mcdonnnj: I wasn't sure if order mattered as far as Terraform was concerned and the targets in other deploy scripts are not fully sorted either so I held back (but had the same thought).

dav3r: Terraform does not care about the order of the targets since it figures out its own order of operations for the resources it needs to deal with.

mcdonnnj: Please see [c69ce41](https://github.com/cisagov/cyhy_amis/pull/594/commits/c69ce41f616fbef9c1b9d6884782dcd0a52ca213).

dav3r: Great, thanks!

PR #613, opened 2023-01-31
Update Ansible role metadata

mcdonnnj:
#  #

## 🗣 Description ##

This pull request updates the role metadata for the Ansible roles defined in this repository. It also adds explicit collection dependencies to the `ansible/requirements.yml` file.

This updates these roles to more closely match the role metadata as found in our stand-alone projects ([cisagov/skeleton-ansible-role](https://github.com/cisagov/skeleton-ansible-role)). This includes the corrections in https://github.com/cisagov/skeleton-ansible-role/pull/115.

## 🧪 Testing ##

Automated tests pass.

### DIFF
--- a/ansible/roles/client_cert_update/meta/main.yml
+++ b/ansible/roles/client_cert_update/meta/main.yml
@@ -6,7 +6,7 @@ galaxy_info:
   galaxy_tags: []
   license: CC0
   # Our standalone Ansible roles require this Ansible version
-  min_ansible_version: 2.10
+  min_ansible_version: "2.10"
   namespace: cyhy
   platforms:
     - name: Debian
--- a/ansible/roles/cyhy_dashboard/meta/main.yml
+++ b/ansible/roles/cyhy_dashboard/meta/main.yml
@@ -6,7 +6,7 @@ galaxy_info:
   galaxy_tags: []
   license: CC0
   # Our standalone Ansible roles require this Ansible version
-  min_ansible_version: 2.10
+  min_ansible_version: "2.10"
   namespace: cyhy
   platforms:
     - name: Debian
--- a/ansible/requirements.yml
+++ b/ansible/requirements.yml
@@ -1,4 +1,9 @@
-- src: https://github.com/cisagov/ansible-role-cyhy-logrotate
-  name: cyhy_logrotate
-- src: https://github.com/cisagov/ansible-role-mongo-logrotate
-  name: mongo_logrotate
+collections:
+  - community.general
+  - community.mongodb
+
+roles:
+  - src: https://github.com/cisagov/ansible-role-cyhy-logrotate
+    name: cyhy_logrotate
+  - src: https://github.com/cisagov/ansible-role-mongo-logrotate
+    name: mongo_logrotate
--- a/ansible/roles/code_gov_update/meta/main.yml
+++ b/ansible/roles/code_gov_update/meta/main.yml
@@ -6,7 +6,7 @@ galaxy_info:
   galaxy_tags: []
   license: CC0
   # Our standalone Ansible roles require this Ansible version
-  min_ansible_version: 2.10
+  min_ansible_version: "2.10"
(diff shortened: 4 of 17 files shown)


ansible/roles/cyhy_dashboard/meta/main.yml
```diff
@@ -15,4 +15,6 @@ galaxy_info:
         - buster
   role_name: cyhy_dashboard
 
+collections:
+  - community.general
```
dav3r: Why is this `collections` block only in this particular `meta/main.yml` file?

dav3r (approved):
Just one question...

ansible/roles/cyhy_dashboard/meta/main.yml
mcdonnnj: I only added `collections` blocks to the metadata files of the two roles that use non-builtin functionality. This role (`cyhy_dashboard`) uses `community.general.make` and the `mongo` roles has it because it uses `community.mongodb.mongodb_user`. I'm happy to add empty `collections: []` entries to all of the other roles for consistency.

dav3r: I'm good with adding empty collections entries for consistency.  @jsf9k, you good with that?

dav3r: Maybe also include a standard explanatory comment for those (like me) who were not familiar with the reasoning there.

jsf9k: > I'm good with adding empty collections entries for consistency. @jsf9k, you good with that?

Yes, I am good with this.  We should also add it to [`cisagov/skeleton-ansible-role`](https://github.com/cisagov/skeleton-ansible-role).

ansible/roles/client_cert_update/meta/main.yml
```diff
@@ -1,12 +1,12 @@
 ---
 galaxy_info:
   author: VM Fusion Dev
-  company: CISA Cyber Assessments
   description: Set up the requirements for the cisagov/client-cert-update Docker image
+  company: CISA Cyber Assessments
```
jsf9k: Why are `company` and `description` being explicitly moved out of alphabetical order?

jsf9k (changes requested):

ansible/roles/client_cert_update/meta/main.yml
mcdonnnj: That is the order in [cisagov/skeleton-ansible-role](https://github.com/cisagov/skeleton-ansible-role/blob/95681aa71c17c7cf69d4658810547dec13d60ccb/meta/main.yml#L3-L5) and part of this PR is making these roles align more with our independent Ansible roles.

jsf9k: I think the order should be fixed in `cisagov/skeleton-ansible-role` then, and can be left alone here.

mcdonnnj: I have removed that group of changes as part of my rebase.

ansible/roles/cyhy_dashboard/meta/main.yml
mcdonnnj: So coming back here I actually found official documentation that had me remove these entries. Per the [documentation](https://docs.ansible.com/ansible/latest/collections_guide/collections_using_playbooks.html#using-collections-in-roles) that keyword is only used for namespace searching and is not involved in dependency resolution. Per https://github.com/ansible/ansible/issues/76030#issuecomment-942520399 there is no way for a role to depend on a collection as part of role metadata. The only way around this would be converting these Ansible roles into a collection and defining any collection dependencies there.

dav3r: Even better- thanks for looking into that. 👍

dav3r (approved):

jsf9k (approved):

PR #640, opened 2023-03-28
Move `cyhy` user creation to AMI build time

### LINKED ISSUE #545: Set up the `cyhy` user and group using an Ansible role at AMI build time
# 🚀 Feature Proposal #

We should switch to setting up the `cyhy` user and group as part of the AMI build with Packer.

## Motivation ##

Currently we set up the `cyhy` user and group [using cloud-init](https://github.com/cisagov/cyhy_amis/blob/develop/terraform/cloud-init/cyhy_user_ssh_setup.yml). We should switch to creating this user as one of the first steps in the AMI build process since it is used throughout the configuration of these images.

## Pitch ##

This aligns the AMIs built in this configuration more closely with the ones in the COOL system. It will also allow us to correctly set ownership of files created in later Ansible roles as needed.

## Implementation notes ##

This will require a new Ansible role (something like `ansible-role-cyhy-user`). We will also need to implement adding the SSH public key as part of that role as it is also set up by cloud-init.

mcdonnnj:
#  #

## 🗣 Description ##

This pull request moves creation of the `cyhy` user to AMI build time from being part of individual instance cloud-init configurations.

This change moves user creation to AMI build time to provide a better guarantee that the `cyhy` user exists before any Ansible provisioners are run against an instance during deployment. It will also allow the option for cyhy-specific Ansible roles to expect the existence of the `cyhy` user as part of their logic. As a bonus this reduces complexity by removing this element of the individual cloud-init configurations for instances that need this user.

## 🧪 Testing ##

Automated tests pass. I used this branch in my test environment and verified that the `cyhy` user existed and that SSH access worked as expected.

### DIFF
--- a/packer/ansible/create_cyhy_user.yml
+++ b/packer/ansible/create_cyhy_user.yml
@@ -21,5 +21,5 @@
 
     - name: Add the SSH public key as an authorized key
       ansible.posix.authorized_key:
-        user: cyhy
         key: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOreUDnms12MPI0gh7K+YGaESYgC2TY1zA+kSK/g+n5+ cyhy
+        user: cyhy
--- a/packer/ansible/create_cyhy_user.yml
+++ b/packer/ansible/create_cyhy_user.yml
@@ -1,0 +1,25 @@
+---
+# The bastion is the only instance that does not need a cyhy user for
+# operational functionality.
+- hosts: all:!bastion
+  name: Create the cyhy user and set up SSH access
+  become: yes
+  become_method: sudo
+  tasks:
+    - name: Create the cyhy user
+      ansible.builtin.user:
+        home: /var/cyhy
+        name: cyhy
+        shell: /bin/bash
+        uid: 2048
+      register: user_info
+
+    - name: Modify permissions on the home directory
+      ansible.builtin.file:
+        mode: 0750
+        path: "{{ user_info.home }}"
+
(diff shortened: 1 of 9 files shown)


jsf9k (approved):

packer/ansible/create_cyhy_user.yml
```diff
@@ -0,0 +1,25 @@
...
+
+    - name: Modify permissions on the home directory
+      ansible.builtin.file:
+        mode: 0750
+        path: "{{ user_info.home }}"
+
+    - name: Add the SSH public key as an authorized key
+      ansible.posix.authorized_key:
+        user: cyhy
+        key: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOreUDnms12MPI0gh7K+YGaESYgC2TY1zA+kSK/g+n5+ cyhy
```
dav3r: If we want to maintain alphabetical order.  If not, I understand. 😄 
```suggestion
        key: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOreUDnms12MPI0gh7K+YGaESYgC2TY1zA+kSK/g+n5+ cyhy
        user: cyhy
```

dav3r (approved):
👍 with one minor thang that may be ignored if you so choose.

packer/ansible/create_cyhy_user.yml
mcdonnnj: Ah, nope I missed that while migrating the configuration. Glad you caught it!

PR #657, opened 2023-05-16
Update `boto3` and `botocore` installation on Debian Buster

mcdonnnj:
#  #

## 🗣 Description ##

This pull request updates the logic for Debian Buster-based AMIs to install [boto3] and [botocore] directly into the Python 3 environment instead of as Debian system packages.

We need a newer version of [botocore] to support [IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html). We were installing the version of `python3-botocore` from Debian Buster Backports, but the version of `python3-boto3` available for Debian Buster is too old in comparison as there is no version available from Backports. This causes a failure for the [cisagov/cyhy-feeds] project as it is Python 3 and requires both packages. We resolve this by ensuring the system packages are not installed, installing the Python versions directly, and then continuing with the installation of the `awscli` from Debian Buster Backports.

## 🧪 Testing ##

Automated tests pass. I used this in my development environment and verified that both `awscli` and the [cisagov/cyhy-feeds] project worked as expected.

[boto3]: https://pypi.org/project/boto3/
[botocore]: https://pypi.org/project/botocore/
[cisagov/cyhy-feeds]: https://github.com/cisagov/cyhy-feeds

### DIFF
--- a/packer/ansible/aws.yml
+++ b/packer/ansible/aws.yml
@@ -11,16 +11,31 @@
     # volumes as NVMe block devices, so that's why we need nvme here.
     - nvme
   tasks:
-    # The versions of awscli and python3-botocore in Debian Buster are too
-    # old to support IMDSv2. There are versions published to buster-backports
-    # that do support IMDSv2, so we should use them.
-    - name: Install AWS packages from buster-backports for Debian Buster
-      ansible.builtin.apt:
-        default_release: buster-backports
-        name:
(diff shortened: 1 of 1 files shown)


jsf9k (approved):

dav3r (approved):
:shipit:

PR #689, opened 2023-10-07
Update Ansible requirements files

mcdonnnj:
#  #

## 🗣 Description ##

This pull request updates the `requirements.yml` files for the post-deployment Ansible configuration in `ansible/` and the Packer Ansible configuration in `packer/ansible/`.

I was doing some testing with the latest Ansible version and ran into a dependency issue due to the effort to reduce what modules are bundled with Ansible. Any modules that aren't in the `ansible.builtin` collection should be listed as a requirement to ensure all dependencies are installed.

> [!NOTE]
> This required changing the format of the `requirements.yml` file for the Packer Ansible configuration so that a collection could be listed as a requirement.

## 🧪 Testing ##

Automated tests pass.

### DIFF
--- a/ansible/requirements.yml
+++ b/ansible/requirements.yml
@@ -1,4 +1,5 @@
 collections:
+  - ansible.posix
   - community.general
   - name: community.mongodb
     # The last version that officially supports MongoDB 3.6
--- a/packer/ansible/requirements.yml
+++ b/packer/ansible/requirements.yml
@@ -1,76 +1,80 @@
-- name: amazon_ssm_agent
-  src: https://github.com/cisagov/ansible-role-amazon-ssm-agent
-- name: apt_over_https
-  src: https://github.com/cisagov/ansible-role-apt-over-https
-- name: automated_security_updates
-  src: https://github.com/cisagov/ansible-role-automated-security-updates
-- name: backports
-  src: https://github.com/cisagov/ansible-role-backports
-- name: banner
-  src: https://github.com/cisagov/ansible-role-banner
-- name: chrony_aws
-  src: https://github.com/cisagov/ansible-role-chrony-aws
-- name: clamav
-  src: https://github.com/cisagov/ansible-role-clamav
-- name: client_cert_update
-  src: https://github.com/cisagov/ansible-role-client-cert-update
-- name: cloudwatch_agent
-  src: https://github.com/cisagov/ansible-role-cloudwatch-agent
-- name: code_gov_update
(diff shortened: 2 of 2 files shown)


jsf9k (approved):

dav3r (approved):
Thanks for keeping us current! 👍

packer/ansible/requirements.yml
```diff
@@ -1,76 +1,80 @@
...
-- name: remove_python2
-  src: https://github.com/cisagov/ansible-role-remove-python2
-- name: upgrade
-  src: https://github.com/cisagov/ansible-role-upgrade
-- name: vdp_scanner
-  src: https://github.com/cisagov/ansible-role-vdp-scanner
-- name: xfs
-  src: https://github.com/cisagov/ansible-role-xfs
+collections:
+  - ansible.posix
```
jsf9k, line 2: If you know what modules you are using from `ansible.posix` you might think about adding a `version` key here [à la cisagov/ansible-role-vnc-server](https://github.com/cisagov/ansible-role-vnc-server/blob/207d8d52588bbc59f2dabfb1e021bb00344ec80d/meta/requirements.yml#L12-L17).

ansible/requirements.yml
```diff
@@ -1,4 +1,5 @@
 collections:
+  - ansible.posix
   - community.general
```
jsf9k, line 3: If you know what modules you are using from `ansible.posix` and `community.general` then you might think about adding a `version` key here [à la cisagov/ansible-role-vnc-server](https://github.com/cisagov/ansible-role-vnc-server/blob/207d8d52588bbc59f2dabfb1e021bb00344ec80d/meta/requirements.yml#L12-L17)

jsf9k (commented):
A suggestion that you can take or leave.

packer/ansible/requirements.yml
mcdonnnj: When I looked into this (I had the same thought) the module used was included in the 1.0 release with the options we are using. As a result I decided not to worry about adding a `version` key.

ansible/requirements.yml
mcdonnnj: When I looked into this (I had the same thought) the module from `ansible.posix` used was included in the 1.0 release with the options we are using. As a result I decided not to worry about adding a `version` key. It appears that the module used from `community.general` with the options we are using has been present since before it was broken out of being included with Ansible.

PR #738, opened 2023-12-26
Update the security header Lambda's runtime to Node 18

jsf9k:
## 🗣 Description ##

This pull request updates the security header Lambda's runtime from Node 16 to Node 18.

## 💭 Motivation and context ##

We recently received an email from AWS Lambda stating that they are dropping support for Node 16 in mid-2024.  Now that we are using version 4.9 of the Terraform AWS provider, we can upgrade our Lambda runtime to Node 18.  When we move to the latest version of the Terraform AWS provider, we will be able to upgrade the runtime to at least Node 20.

## 🧪 Testing ##

I tested this in the most brütal fashion - by deploying it to production.

## ✅ Pre-approval checklist ##

- [x] This PR has an informative and human-readable title.
- [x] Changes are limited to a single goal - *eschew scope creep!*
- [x] All relevant type-of-change labels have been added.
- [x] I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document.
- [x] These code changes follow [cisagov code standards](https://github.com/cisagov/development-guide).
- [x] All new and existing tests pass.

PR #654, opened 2023-05-15
Break out S3 bucket settings using backported resources

mcdonnnj:
#  #

## 🗣 Description ##

This pull request breaks out settings for AWS S3 buckets into their own resources. In some cases additional resources were added around encryption and/or public access blocking.

## 💭 Motivation and context ##

In the 4.0 release of the Terraform AWS provider they performed a refactor of the overloaded S3 resource per [this post](https://www.hashicorp.com/blog/terraform-aws-provider-4-0-refactors-s3-bucket-resource). In the 3.75.0 release they [backported these S3 changes](https://github.com/hashicorp/terraform-provider-aws/blob/release/3.x/CHANGELOG.md#3750-march-18-2022). This allows us to make these changes before migrating to the 4.x release of this provider (and is the only known breaking change).

## 🧪 Testing ##

Automated tests pass. I deployed this in my development environment without issue.

## ✅ Pre-approval checklist ##

- [x] This PR has an informative and human-readable title.
- [x] Changes are limited to a single goal - *eschew scope creep!*
- [x] All relevant type-of-change labels have been added.
- [x] I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document.
- [x] These code changes follow [cisagov code standards](https://github.com/cisagov/development-guide).
- [x] All relevant repo and/or project documentation has been updated
      to reflect the changes in this PR.
- [x] All new and existing tests pass.

PR #742, opened 2023-12-28
Fix the rules publication Terraform configuration

mcdonnnj:
#  #

## 🗣 Description ##

This pull request updates the configuration in `terraform_egress_pub/` that publishes the [rules site] listing the IP ranges we publish that are in use for assessments. This includes:
- Removing the S3 static website configuration.
- Adding an OAC ([Origin Access Control]) for the CloudFront distribution.
- Adding a policy to the rules S3 bucket to allow read access from the OAC now associated with the CloudFront distribution.

When #654 was merged I apparently neglected to apply the changes in the `terraform_egress_pub/` configuration. As a result a bug went unnoticed until @jsf9k applied his changes for #738 and suddenly the [rules site] was no longer working. I took a look and realized that the problem was due to the fact that although the bucket was set to use the `private` canned ACL, the script that uploads objects to the bucket would set each uploaded object to use the `public-read` canned ACL. With the addition of a `aws_s3_bucket_public_access_block` for the rules S3 bucket that disallowed _any_ public access it broke access to the objects in the bucket. Adding an OAC and appropriate policy to the bucket allows the CloudFront distribution to access the S3 bucket even if it is completely private.

The static website configuration was removed because
1) The bucket is now completely private.
2) The CloudFront distribution handles the same functionality making it superfluous. 

## 🧪 Testing ##

Automated tests pass. I applied this in production and verified that the [rules site] works once more.

[Origin Access Control]: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html
[rules site]: https://rules.ncats.cyber.dhs.gov

### DIFF
--- a/terraform_egress_pub/cloudfront.tf
+++ b/terraform_egress_pub/cloudfront.tf
@@ -69,18 +69,20 @@ module "security_header_lambda" {
   tags                   = { "Application" = "Egress Publish" }
 }
 
-resource "aws_cloudfront_origin_access_identity" "rules_s3_distribution" {
-  comment = var.distribution_oai_comment
+resource "aws_cloudfront_origin_access_control" "rules_s3_distribution" {
+  description = var.distribution_oac_description
+  name        = var.distribution_oac_name
+
+  origin_access_control_origin_type = "s3"
+  signing_behavior                  = "always"
+  signing_protocol                  = "sigv4"
 }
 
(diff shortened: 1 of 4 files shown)


jsf9k (approved):
Thank you for fixing this!

dav3r (approved):
👍 👍  Thanks for dragging us into the future, kicking and screaming all the way!

PR #754, opened 2024-04-17
Remove suffix from the argument to the `sleep` command

mcdonnnj:
#  #

## 🗣 Description ##

This pull request removes the suffix indicating the unit of time for the `sleep` command in the two `local-exec` commands used to set up PTR records for vulnscan instances.

The version of `sleep` available on macOS does not support suffixes and only accepts a number of seconds to sleep. This appears to have changed in a recent macOS release as these `local-exec`s ran without issue before. The other possibility is that the Homebrew `coreutils` package was not prefixing the GNU `sleep` command previously and since they currently are it no longer overrides the macOS default. Either way it does not work and so we must rectify that.

## 🧪 Testing ##

Automated tests pass. I needed this change to successfully deploy my testing environment with no issues.


jsf9k (approved):

dav3r (approved):
⏱️

PR #858, opened 2025-05-21
Use symbolic permissions in Ansible code

mcdonnnj:
#  #

## 🗣 Description ##

This pull request replaces all uses of implicit permission (octal permissions representation) with symbolic permissions.

In addition to being easier to read this will also be required by the linter changes in #856. Since there is nothing stopping us from doing it "in advance" and due to the amount of Ansible code that lives in this repository I thought it would be helpful to break the change out into its own pull request.

## 🧪 Testing ##

Automated tests pass. I verified that no implicit permissions remain with

```console
$ grep -rnE ': 0[0-7]{3}'
```

### DIFF
--- a/ansible/roles/code_gov_update/tasks/main.yml
+++ b/ansible/roles/code_gov_update/tasks/main.yml
@@ -5,7 +5,7 @@
 - name: Create the secrets directory
   ansible.builtin.file:
     group: cyhy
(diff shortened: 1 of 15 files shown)


jsf9k (approved):
So proactive!

dav3r (approved):

PR #972, opened 2026-08-31
Increase MongoDB journal EBS volume size from 8 GiB to 16 GiB

dav3r:
#  #

## 🗣 Description ##

This PR increases the MongoDB journal EBS volume size from 8 GiB to 16 GiB.

Our current snapshot creation process today filled up the 8 GiB journal volume, so we are doubling it to give us more breathing room.  I suspect that this is due to recent changes which allow us to create all of our snapshots in a much shorter amount of time and our MongoDB instance is not able to keep up with the increased volume of writes.

In the near future, I plan to modify the snapshot/report script to confirm that all of the snapshots have been written to the DB before starting to generate reports.

## 🧪 Testing ##

Visual inspection.

I have already applied this change in Production, so the proof that this was helpful will come during the next reporting cycle.  If MongoDB doesn't crash like it did this week, we will consider this a success.


mcdonnnj (approved):
:shipit:

jsf9k (approved):

